PR #6, opened 2014-01-09
Miscellaneous upgrades to the slider.

nadavsamet:
* Slider and module renaming to be more generic.
* Slider attributes renaming.
* Label positioning fix to stay inside the bar's bounds.
* Minor graphical changes.
* Slider can now be used as an attributes to simplify support in IE.


coomsie:
grt work, should be merged! Only one that works for me. :P

nadavsamet:
@coomsie Thanks!

PR #27, opened 2014-09-12
Class prefixes

ropsu:
This project uses slightly generic class names. This PR adds `rz-` prefix to class names. This decreases the likelihood of accidentally applying some external styles to the slider because of class name collisions.


rzajac:
Merged in

PR #32, opened 2014-10-29
Updated to work with jqLite/Ionic

willyboy:
Bound touchmove/mousemove to slider instead of document. Used $document.one to remove event on touchend/mouseup

### DIFF
--- a/rzslider.js
+++ b/rzslider.js
@@ -4,7 +4,7 @@
  * (c) Rafal Zajac <rzajac@gmail.com>
  * http://github.com/rzajac/angularjs-slider
  *
- * Version: v0.1.6
+ * Version: v0.1.6r
  *
  * Licensed under the MIT license
  */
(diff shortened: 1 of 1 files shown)


leandroz:
merge?

Yuripetusko:
+1 for the merge

Yuripetusko:
the only thing is missing here is bower main files are not properly defined. So my gulp wiredep cannot pick correct files to inject in html

it should be something like:

    "main": [
        "rzslider.js",
        "dist/rzslider.min.css"
    ],

Yuripetusko:
I doesn't look like you merged it properly? For example I cannot see `this.sliderElem.on('touchmove',` code in your master now

rzajac:
Unfortunately I had to go back to binding events to $document due to some problems - but I forgot to document it in the commit :(

It happened here: 3772b558af8bfe5f50ae0e8978c12b3487b83303

Yuripetusko:
Ok. Well it doesn't work with Ionic this way :/

Yuripetusko:
maybe @willyboy can help out.?

willyboy:
I'll take a look tonight. @rzajac Any specific problems I might be able to address in a fix?

rzajac:
If we could find a way for it to work in Ionic that would be great. I don't use it so I have no way to test it.

PR #36, opened 2014-12-16
progressive step

limpbrains:
Hi. I've added option called "progressive step" if it is true step will be calcualted automatically depending for value.

For example:
if (1 < value < 60 ) step = 1;
if (60 < value < 600) step = 10;
and so on

This is usefull for big values.


rzajac:
This would prevent setting some values. It's better to leave setting step to user.

PR #40, opened 2015-02-12
support ngChange

keithbox:
i use this range selection for a table filter, ng change is really in need!


rzajac:
Looks like this repo no longer exists.

PR #44, opened 2015-04-19
Update rzslider.less to contemplate all colors

fabiowitt:


rzajac:
The new LESS file was added in the latest release

PR #45, opened 2015-05-11
change "Slider" service to "RzSlider" to avoid collision

ptgamr:
Since "Slider" is quite a common name, and there is possibility that this name got collision with other code. I got the collision with this `angular-bootstrap-colorpicker`. We should rename it to something more unique. I could think of `RzSlider` in this case.


ptgamr:
Thanks @rzajac !

Issue #26, opened 2014-09-06
add selection bar when slider have only one pointer


corcisun:
Now the selection bar "<span class="bar selection"></span>" is rendered only when I define two pointers, such as the first example in your plunker (http://embed.plnkr.co/53AUdB/preview). 
 
It's better if the selection bar appears when I want, also when i use only one pointer..
You could set the selection bar in slider only by configuration or whatever you want.

If you make this issue you make me happy :) 

thanks.

PS: I can't send a screenshot of what I am thinking because github have some problem with uploads. I will upload an image as soon as i can.

psalaberria002:
+1

PR #48, opened 2015-05-23
Selection bar updates correctly now. 

psalaberria002:
Bug: Selection bar not updating when this.maxH.rzsl returns NaN. Related to #26

### DIFF
--- a/rzslider.js
+++ b/rzslider.js
@@ -653,7 +653,7 @@ function throttle(func, wait, options) {
      */
     updateSelectionBar: function()
     {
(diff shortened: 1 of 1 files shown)


rzajac:
Can you tell me how I can reproduce it? I believe this should be fixed somewhere else.

psalaberria002:
Yes, this happens when adding the slider into a modal (bootstrap modal). I also got the slider initially at zero, but solved that with the refresh event.

I confirm that the exact same slider outside of the modal works perfect.

rzajac:
I have no way to check it now but I believe I fixed it with making sure maxH.rzsl is never NaN

PR #50, opened 2015-05-24
Selection bar not updating after rzSliderForceRender (fixed)

psalaberria002:
After some debugging I found the problem. rzSliderForceRender was calling thrHigh() even when not using range slider.

I guess this is a better fix than the previous one (#48)

### DIFF
--- a/rzslider.js
+++ b/rzslider.js
@@ -260,7 +260,7 @@ function throttle(func, wait, options) {
       {
         self.resetLabelsValue();
         thrLow();
-        thrHigh();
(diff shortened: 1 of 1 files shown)


rzajac:
Will it also work when rz-slider-always-show-bar is set ?

Calling thrHigh() calls updateSelectionBar() which updates selection bar.

psalaberria002:
I am using it with rz-slider-always-show-bar="true", and it works as expected. Since thrLow() calls updateSelectionBar() that's enough.

Issue #49, opened 2015-05-23
Ability to pass the template as a string or a templateUrl for the slider


psalaberria002:
It would be great to be able to edit the template for the slider. Maybe an attribute in the directive? What do you think?

psalaberria002:
What about something like this? I need to change the html inside the pointer, and this solves the issue.

Any other idea?

Thank you.

.directive('rzslider', ['$compile','RzSlider', function($compile,Slider)
{
  return {
    restrict: 'E',
    scope: {
      rzSliderFloor: '=?',
      rzSliderCeil: '=?',
      rzSliderStep: '@',
      rzSliderPrecision: '@',
      rzSliderModel: '=?',
      rzSliderHigh: '=?',
      rzSliderTranslate: '&',
      rzSliderHideLimitLabels: '=?',
      rzSliderAlwaysShowBar: '=?',
      rzSliderTemplate: '=?'
    },

    link: function(scope, elem, attr)
    {
        var template ='<span class="rz-bar"></span>' + // 0 The slider bar
                    '<span class="rz-bar rz-selection"></span>' + // 1 Highlight between two handles
                    '<span class="rz-pointer"></span>' + // 2 Left slider handle
                    '<span class="rz-pointer"></span>' + // 3 Right slider handle
                    '<span class="rz-bubble rz-limit"></span>' + // 4 Floor label
                    '<span class="rz-bubble rz-limit"></span>' + // 5 Ceiling label
                    '<span class="rz-bubble"></span>' + // 6 Label above left slider handle
                    '<span class="rz-bubble"></span>' + // 7 Label above right slider handle
                    '<span class="rz-bubble"></span>'; // 8 Range label when the slider handles are close ex. 15 - 17
                    
        if(scope.rzSliderTemplate !== undefined){
            template=scope.rzSliderTemplate;
        }
        elem.html(template);
        $compile(elem.contents())(scope);

        return new Slider(scope, elem, attr);
    }
  };
}]);

rzajac:
I would rather go somewhere in this direction: http://stackoverflow.com/a/20659860/282882

rzajac:
Implemented by #51

PR #51, opened 2015-05-24
Ability to pass the template for the slider #49

psalaberria002:
I have done some research regarding #49, and as far as I know there is not way to access the scope in the compile function. Same happens for the template function. You can only access data in $scope inside the LinkingFunction. 

Let me know if you would like to add this feature. I need it for a current project I am working on, so it would be perfect if I could fetch it directly from your repo.

Thank you.

### DIFF
--- a/rzslider.js
+++ b/rzslider.js
@@ -960,7 +960,7 @@ function throttle(func, wait, options) {
   return Slider;
 }])
 
-.directive('rzslider', ['RzSlider', function(Slider)
+.directive('rzslider', ['$compile','RzSlider', function($compile,Slider)
 {
   return {
     restrict: 'E',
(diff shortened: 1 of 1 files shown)


rzajac:
I decided to allow custom templates via templateUrl. See commit 19d51a94bbcfe79aa5065c25f31b33a649efb47e.

psalaberria002:
One suggestion regarding this. Instead of having the default template in a separate file, you should add it to the templateCache in the run section of the module. Then you avoid the 404 error in http://<baseurl>/rzSliderTpl.html.

    module('rzModule', []).run(function($templateCache) {
        var template = '<span class="rz-bar"></span>' + // 0 The slider bar
                '<span class="rz-bar rz-selection"></span>' + // 1 Highlight between two handles
                '<span class="rz-pointer"></span>' + // 2 Left slider handle
                '<span class="rz-pointer"></span>' + // 3 Right slider handle
                '<span class="rz-bubble rz-limit"></span>' + // 4 Floor label
                '<span class="rz-bubble rz-limit"></span>' + // 5 Ceiling label
                '<span class="rz-bubble"></span>' + // 6 Label above left slider handle
                '<span class="rz-bubble"></span>' + // 7 Label above right slider handle
                '<span class="rz-bubble"></span>'; // 8 Range label when the slider handles are close ex. 15 -     17
      $templateCache.put('rzSliderTpl.html', template);
    })

psalaberria002:
Implemented in #53. You can close this one.

PR #52, opened 2015-05-25
Added bootstrap modal to the examples

psalaberria002:


psalaberria002:
I updated it to the latest version, and also added a full example of how to use the slider into a modal.

PR #55, opened 2015-05-25
Hide elements using display none instead of opacity zero

### LINKED ISSUE #54: Drag event not working when using selection bar and not multirange
When dragging the normal slider to the left completely, it stops working. The "hidden" highPointer ruins it apparently. hideEl() should return element.css({display: 'none'}) for the highPointer instead of opacity 0. Otherwise a z-index lower than for lowPointer. 

Not sure what solution is the best.

psalaberria002:
Fixes #54

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -30,7 +30,7 @@
                 rz-slider-model="priceSlider.min"
                 rz-slider-high="priceSlider.max"
                 rz-slider-step="1"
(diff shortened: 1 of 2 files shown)


psalaberria002:
Ok, I just realized that this breaks the max and min labels. :( z-index should be enough?

rzajac:
Fixed by 10011b376ac7f475888fe1b4c4d95bf68f195d06

PR #67, opened 2015-06-05
Add a rzSliderOnChange attribute. 

ValentinH:
```html
<rzslider rz-slider-model="priceSlider" rz-slider-on-change="onSliderChange()"></rzslider>
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,10 @@ $scope.priceSlider = {
 
 > Custom translate function. Use this if you want to translate values displayed on the slider. For example if you want to display dollar amounts instead of just numbers do this:
 
+**rz-slider-on-change**
+
+> Function to be called when rz-slider-model or rz-slider-high change.
+
 ```javascript
 // In your controller
 
@@ -128,7 +132,8 @@ $scope.translate = function(value)
     rz-slider-ceil="priceSlider.ceil"
     rz-slider-model="priceSlider.min"
     rz-slider-high="priceSlider.max"
-    rz-slider-translate="translate"></rzslider>
+    rz-slider-translate="translate"
+    rz-slider-on-change="onSliderChange()"></rzslider>
 ```
 
 ## Slider events
--- a/rzslider.js
+++ b/rzslider.js
@@ -561,6 +561,17 @@ function throttle(func, wait, options) {
     },
 
     /**
+     * Call the onChange callback if defined
+     *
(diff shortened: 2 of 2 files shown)


rzajac:
Similar functionality is already implemented with using `slideEnded ` event: https://github.com/rzajac/angularjs-slider#slider-events

ValentinH:
I disagree because it fires only on moveEnd. If I want to update something everytime the value change, I couldn't. Now with this function, it is possible.

Maybe we can replace my implementation by firing a slideMoved event.

rzajac:
How does it differ from watching the model with scope.$watch() ?

ValentinH:
The result is the same but it is better to avoid $watches. This is why there is a ng-change on default inputs. My goal is to provide a ng-change like feature.

antoescu:
\+ 1 Better than $watch

radek-anuszewski:
I think ng-change could be pretty useful.
Imagine a situation:
I have 5 sliders, each value is displayed to the user, for example in blue circle. When user touch/move Slider, his circle turns into green color - to show user that he set value. Or another use case - for some of Sliders **additional work has to be done**, like in (*real-world app use case*) [Glycemia](https://en.wikipedia.org/wiki/Blood_sugar), where [values] (https://en.wikipedia.org/wiki/Blood_sugar#Normal_values_in_humans) can be in **two units: mmol/L or mg/dL**, where is natural to show 2 **separate Sliders for each unit.** - if we don't know which values user will choose we have to do recalculations between .mmol/L and mg/dL.

By the way, I see "This branch is up-to-date with the base branch", was ng-change added to master branch? I can't see it now, maybe I misunderstood this message.

ValentinH:
Nope it is not yet merged. For now, there are 2 people that want it merged (3 with me) and one person @rzajac that doesn't really want to merge it. Let's see if there are more people that request it...

cosme-benito:
> How does it differ from watching the model with scope.$watch() ?

Well, it think this approach is much more viable when I have a variable list of objects which have a range. With this pull I am able to use simple markup to achieve my goals, while with the watch I would have to manage the contents of my list and add/remove watchers accordingly.

radek-anuszewski:
So @ValentinH , 4 to 1 :)

ValentinH:
OK let's merge it ;)

PR #68, opened 2015-06-05
Add a rzSliderShowTicks

ValentinH:
```js
<rzslider rz-slider-model="priceSlider" rz-slider-show-ticks="true"></rzslider>
```

It displays a tick for each possible value. See the Example on the demo page or on [this fiddle](http://jsfiddle.net/ngrq2kdm/).


soee:
I'm not using it atm. but it is a nice feature, so: +1

adgoncal:
It is a good idea, but I was trying it out on the fiddle and when there are ticks, you can no longer click on the timeline bar and have the handle move to that point. However, clicking above the tick still works.

Also, is a table really the best option for displaying ticks?

ValentinH:
Indeed the ticks should be under the bar that's why. I'll rebase it and fix it.

About the table, why not?

adgoncal:
I just thought there may be a better approach because it is not tabular data. 
But it is simple and it works, so I guess it should be fine.

ValentinH:
Actually, the trick with the table is that you don't need to specify each tick's position. Just give the total width and they will be equally spread.

ValentinH:
@adgoncal rebased so now the ticks are under the bar.

adgoncal:
Cool!

ValentinH:
Actually, the limit with the current table method is that if you want to draw circle as ticks, it is not possible. Indeed, it's using td borders...

ValentinH:
I have replaced the table method by a method with ul displayed with flex. Now you can use whatever you want for the ticks: even circles ;)

See http://jsfiddle.net/ngrq2kdm/

PR #69, opened 2015-06-05
Support clicks on bar

ValentinH:
Add the possibility to directly click on the slider bar to set the value. It is also possible to click and drag directly from the bar.

Add a bar wrapper to enlarge the clickable area.


psalaberria002:
+1 Nice feature.

mcrumm:
+1 This would be super helpful.

antoescu:
+ 1 It is needed

radek-anuszewski:
I checked pull request submitted by @ValentinH on latest Chrome and on Motorola Moto G II with Android 5.0 (by Phonegap Developer app) and everything works well, I think it is worth merge.

PR #75, opened 2015-06-17
Fixes #74: Range Slider - No label if "Min" is the same as "Floor"

### LINKED ISSUE #74: Range Slider - No label if "Min" is the same as "Floor"
__Usage__
I am using the Range Slider of this example:
https://github.com/rzajac/angularjs-slider#range-slider

__Issue__
If minimum value is set to the same value as the floor value, then the value label is not being displayed. When you drag the min value, then value label appears even when dragged all the way to the floor value. So min does not behave as it should, like max for example.

__Reproduce Example__
JS Fiddle: https://jsfiddle.net/tjvuy4z5/1/

m-spyratos:
Before we stop proceeding because of a small delta, we check if min label already exists or not. If it does not exist, even if delta is small, we proceed so min label can get it's initialize value.

### DIFF
--- a/rzslider.js
+++ b/rzslider.js
@@ -609,7 +609,7 @@ function throttle(func, wait, options) {
     {
       var delta = Math.abs(this.minH.rzsl - newOffset);
 
-      if(this.minLab.rzsv && delta <= 0 && delta < 1) { return; }
+      if(delta <= 0 && delta < 1) { return; }
 
       this.setLeft(this.minH, newOffset);
       this.translateFn(this.scope.rzSliderModel, this.minLab);
--- a/rzslider.js
+++ b/rzslider.js
@@ -609,7 +609,7 @@ function throttle(func, wait, options) {
     {
       var delta = Math.abs(this.minH.rzsl - newOffset);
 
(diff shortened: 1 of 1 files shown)


ValentinH:
Actually, I don't really know what is the goal of this code:
```js
if(delta <= 0 && delta < 1) { return; }
```

In my project, I simply removed it to avoid the issue you mentioned in #74.

In your commit, what is the reason why you use `this.minLab.rzsv`?

m-spyratos:
It's been couple of weeks since I made this commit, so I don't remember exactly why I chose this fix. I remember though that the delta check is important and removing it was breaking my project. Depends what type of sliders/options you are using.

How come is wrongly positioned? Did you check the position with my code and without it? Do you have screenshots to share? At least for its working fine with this addition.

ValentinH:
This is the expected result (the one I have by simply commenting the condition and when I have move the slider at least once):
image


This is the result with your commit:
image
The left label is wrongly positioned and when I move the slider and come back to this value, it gets the correct position (the one of the first screenshot).

To finish, for completion sake, this is the result with the current master code:
image

PR #77, opened 2015-06-26
Fix delta checking on handle update

joshrickert:
Very minor, but I assume there was a typo here. A conditional on
```javascript
delta <= 0 && delta < 1
```
makes no sense, because the second statement definitely false if the first is. It seems like you meant to check if the delta was between 0 and 1.


ValentinH:
I guess you're right because it doesn't make sense. However, this code is linked with #74 and we need to find a solution.

PR #80, opened 2015-07-13
Add SASS file

soee:
Converted LESS to SASS styles file.

### DIFF
--- a/rzslider.scss
+++ b/rzslider.scss
@@ -1,0 +1,116 @@
+/**
+ * Angular JS slider directive
+ *
+ * (c) Rafal Zajac <rzajac@gmail.com>
+ * http://github.com/rzajac/angularjs-slider
+ *
+ * Licensed under the MIT license
+ */
+
+/* Slider colors */
+$handleActiveColor: #451aff;
+$handleHoverColor: #fff;
+$labelTextColor: #55637d;
+$handleBgColor: #0db9f0;
+$handleInnerColor: #fff;
+$limitLabelTextColor: $labelTextColor;
+$barFillColor: $handleBgColor;
+$barNormalColor: #d8e0f3;
+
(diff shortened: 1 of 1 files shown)


ValentinH:
I think it is not really good to use 2 different preprocessors because it means that we need to maintain them at each change.

soee:
@ValentinH the styles are very short and difference between aren't really big so maintaining it is pretty easy in my opinion. Having both LESS and SASS would make users life a lot easier :)

ValentinH:
OK, let's wait to see if there are more people voting for this PR. ;)

northamerican:
+1 SASS

kukac7:
:+1:

ValentinH:
I'm closing it since the files architecture has changed a lot since this PR. Also, the .less file has changed.

If SASS and LESS are that similar, then it is easy for a user to convert the less file. It will be a mess to maintain both and only one will be used to generate the dist CSS file...

soee:
Hi, i think you should think about switching from LESS to SASS than. 

Upcoming Twitter Bootstrap 4 CSS framework is moving to SASS: 
"Moved from Less to Sass. Bootstrap now compiles faster than ever thanks to Libsass, and we join an increasingly large community of Sass developers."

If i'm not wrong ZURB Foundation 6 CSS Framework also moves to SASS. They use it for Motion UI:
"In preparation for the upcoming release of Foundation for Sites 6, today we're open-sourcing Motion UI, a Sass library for creating custom CSS transitions and animations."

All the big players moves to SASS - let's do the same imo. :)

ValentinH:
Why not directly using postCSS? :D 

See this post of Bootstrap creator: https://twitter.com/mdo/status/591364406816079873

alexeybondarenko:
+1 Sass

PR #81, opened 2015-07-14
Remove invalid text from README.md

soee:
Inside **Slider events** description there was *demo/tabs.html* file mentioned. As this file does not exist, we should remove this part from README.md.


ValentinH:
Indeed it was wrong, I replaced it with a link on the correct example in this commit e0e5bbe46

PR #83, opened 2015-07-16
Allow the merge of 2 previously incompatible features

mabhub:


ValentinH:
What was the problem with the 2 features? I'm using all the features I submitted through PR together and I don't have any incopatibility. Was it a merge conflict?

PR #92, opened 2015-07-26
show steps label

mahmutduva:
you can show steps label with this option

### DIFF
--- a/rzslider.js
+++ b/rzslider.js
@@ -410,13 +410,15 @@ function throttle(func, wait, options) {
        * @returns {undefined}
        */
     updateTicksScale: function(value, label){
-          var valStr = '';
-          var halfTickWidth = 15;
-          for(var i=0; i<value.length;i++){
+            var valStr = '';
+            var halfTickWidth = 15;
+            for(var i=0; i<value.length;i++){
               var step = (label.rzsw - this.handleHalfWidth * 2) / (value.length - 1);
               valStr += '<li style="left: '+ (((step * i)) - (halfTickWidth - this.handleHalfWidth))   +'px;">' + value[i] + '</li>';
-          }
-          label.html(valStr);
+            }
+            label.html(valStr);
+
+
     },
 
     /**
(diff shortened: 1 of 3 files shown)


ValentinH:
Thanks for this PR, this is a pretty cool feature! Just have a few remarks:

- this does not support using a translate function (if I want to add a '$' before each value for instance)
- this does not scale well (try replacing 600 by 60000 and 150 by 15000 in your example, then the labels are not positioned correctly)
- I don't really like the idea of modifying the `translateFn` like you did, I think this feature need its own function to position eveything (for example, like I did for the [show-ticks PR](https://github.com/rzajac/angularjs-slider/pull/68/files#diff-0539d03e9ce14a40acbd41c92f2caa2eR560) with the `updateTicksScale` function).

rzslider.js
```diff
@@ -392,18 +412,34 @@ function throttle(func, wait, options) {
...
+
       useCustomTr = useCustomTr === undefined ? true : useCustomTr;
 
-      var valStr = String(useCustomTr ? this.customTrFn(value) : value),
-          getWidth = false;
+      if(label[0].className == 'rz-labels'){
+          var valStr = '';
+          console.log(this.barWidth)
+           for(var i=0; i<value.length;i++){
+                valStr += '<li style="left: '+ ((label.rzsw - 38) / (value.length - 1)) * i  +'px">' + value[i] + '</li>';
```
ValentinH: What is 38?

mahmutduva:
tick position bug fix
create new function 'updateTicksScale'

ValentinH:
Indeed, this is much better like this.
Just to be perfect, check this example:
image

The labels are not centered... I don't know if there is a simple way to make them centered.

ValentinH:
Moreover, the browser crashes if I don't provide the step argument!

sebastialonso:
Any update on this PR?

ValentinH:
Nope nothing since my last 2 remarks...

mahmutduva:
I am working about label position. I will commit this update today.

e-cloud:
@mahmutduva , can you build some simple demo to show the feature?

ValentinH:
There you go: http://jsfiddle.net/aq2r7dL8/

ValentinH:
@mahmutduva the labels still don't show if I don't set the step arg...

ValentinH:
I'm closing this PR because I just add this feature on master with `rz-slider-show-ticks-value`.
See http://jsfiddle.net/0xg0bf2a/

PR #96, opened 2015-07-29
Present only with real time update

smariussorin:
The current state of present only is low, does not have real time update.
Added real time update, added disabled on slider for css styling on demand

### DIFF
--- a/rzslider.js
+++ b/rzslider.js
@@ -14,18 +14,18 @@
 
 angular.module('rzModule', [])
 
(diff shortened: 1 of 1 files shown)


ValentinH:
Sorry but I don't understand what it is. Could you provided an example.

Moreover, you have changed the indentation so all the file is modified.Could you please modify your PR to only show what you have changed?

ValentinH:
Closing as no update for more than a month.

Issue #78, opened 2015-07-09
Step is not working using value from scope


phuyem:
Hello,

This one works
    <rzslider
     rz-slider-step="10"
    rz-slider-floor="priceSlider.floor"
    rz-slider-ceil="priceSlider.ceil"
    rz-slider-model="priceSlider.min"
   ></rzslider>

This one doesn't

    <rzslider
     rz-slider-step="priceSlider.step"
    rz-slider-floor="priceSlider.floor"
    rz-slider-ceil="priceSlider.ceil"
    rz-slider-model="priceSlider.min"
   ></rzslider>

Given
$scope.priceSlider = {
    min: 100,
    max: 180,
    ceil: 500,
    floor: 0,
    step:10
};

Thanks for the great plugin

phuyem:
Fix it myself : 
Line 1029
      rzSliderStep: '@',  

=>

      rzSliderStep: '=?',

bdteo:
Yes it doesn't work. Values become NaN and the slider gets stuck.

ValentinH:
Could you please provide a JSFiddle showing the issue?

phuyem:
hey folks, I already gave the fix , didn't I. See the 2nd post. 
Open the source file,

Line 1029
rzSliderStep: '@',

=>

  rzSliderStep: '=?',

Thanks

ValentinH:
I think setting the rzSliderStep as a string attribute was on purpose because it was not supposed to be changed. Maybe for performance reason, the author didn't want to use Two-way model binding.

Changing it may break existing applications using the slider so it has to been think twice...

For now, you can simply use `{{}}` to interpolate your value like this:
```html
<rzslider
 rz-slider-step="{{ priceSlider.step }}"
rz-slider-floor="priceSlider.floor"
rz-slider-ceil="priceSlider.ceil"
rz-slider-model="priceSlider.min"
```

colm-mcqs:
Live saved. Been banging my head against this issue for a day!

ValentinH:
With #158, this problem will be gone.

PR #103, opened 2015-08-13
Allow changing step attribute

northamerican:
The rz-slider-step attribute is static and does not change, as do the min and max. See #78, which was closed without a working solution.

### DIFF
--- a/rzslider.js
+++ b/rzslider.js
@@ -301,6 +301,13 @@ function throttle(func, wait, options) {
 
       // Watchers
 
+      unRegFn = this.scope.$watch('rzSliderStep', function(newValue, oldValue)
+      {
+        if(newValue === oldValue) return;
+        self.step = newValue;
+      });
+      this.deRegFuncs.push(unRegFn);
+
       unRegFn = this.scope.$watch('rzSliderModel', function(newValue, oldValue)
       {
         if(newValue === oldValue) { return; }
@@ -1055,7 +1062,7 @@ function throttle(func, wait, options) {
     scope: {
       rzSliderFloor: '=?',
       rzSliderCeil: '=?',
-      rzSliderStep: '@',
+      rzSliderStep: '=?',
       rzSliderPrecision: '@',
       rzSliderModel: '=?',
       rzSliderHigh: '=?',


ValentinH:
I understand what you mean. However, as I said in #78, this change will probably break existing apps (like if someone is using `rz-slider-step="{{ step }}"`). 

Moreover, when changing the step with your solution, a NaN appears briefly above the slider so I think you are missing something... `self.resetSlider();` maybe?

I think if we do this, we should do the same for the `rz-slider-precision` and other attrs using @ bindings to be more consistent... Then it would mean we bump to a breaking change release so 0.2.0.

Martin-Wegner:
If you don't want to break existing implementations why don't you introduce a new 'rz-slider-nonStaticStep' or something?

ValentinH:
Hmmm I'm not really fan of this solution because I already think that there are too many attributes on this directive.... I'm even considering using a `rz-slider-options` that would use an object containing all the non-mandatory fields and only keep `rz-slider-model` and `rz-slider-high`. Then all the options would be updatable at runtime.

What do you think of this solution? This would definitively be a 2.0 release.

Martin-Wegner:
Sounds good :+1:

Martin-Wegner:
Stupid question... Why does this dynamic step change works?
https://jsfiddle.net/zfee1pxb/3/

ValentinH:
There is no watch on step alone but since you modify other properties that are watched (model for instance), the slider is refreshed and so the step is updated.

Martin-Wegner:
OK and here https://jsfiddle.net/zfee1pxb/5/ the update is triggered when moving the bubble right? So the only thing that does not work yet is that the bubble does not jump around when changing the step right?

ValentinH:
Not only because if you show the ticks on the slider, then they are not updated either:
https://jsfiddle.net/pom92p9r/

ValentinH:
I have started to implement what I mention above about rzSliderOptions:
https://github.com/rzajac/angularjs-slider/blob/2.0.0/src/rzslider.js#L34 .
What do you think?

Martin-Wegner:
It looks like mentioned :)

ValentinH:
Please see #158

PR #123, opened 2015-09-09
Remove the leading parameter of the throttle function calls

ValentinH:
It was adding a lag when the model changes from outside the slider.


jacob-meacham:
+1, this makes the sliders feel a lot less sluggish in some cases.

ValentinH:
It has been done in #153

PR #124, opened 2015-09-12
fix #118

### LINKED ISSUE #118: [refactor request] can I do some refactorization and improvement
there are some point i want to improve:
* add `umd` support
* add *templateCache plugin* to collect the html, it's not nice to write html in `.js` file
* remove the annotation code, and pass it the `ngAnnotate` as for source code doesn't need to be annotated when developping
* [optional] give up `grunt` and be in favor of `gulp`

e-cloud:
fix #118
* add umd support and move the source to the `src` directory
* take the *html fragment* out of the js file, and use *grunt-angular-templates* and *grunt-replace* plugin to generate the templateCache and insert it into the js file
* apply ngAnnotate plugin to the build flow


e-cloud:
one thing is forgot to be mentioned -- provide an unminified version in the dist directory for the src's one is uncompleted

ValentinH:
Thanks for this PR.

PR #128, opened 2015-09-23
Remove unnecessary dependency

a-tarasyuk:
After release 0.1.31 appeared error

Uncaught Error: Cannot find module "underscore"


ValentinH:
You are right. I don't know why @e-cloud added it... Maybe because of the throttle function but @rzajac added it in the code base to remove the underscore dependency.

a-tarasyuk:
@ValentinH could you update version from 0.1.31 to 0.1.32 ?

ValentinH:
it's done. Tell me if it's ok for you please ;)

a-tarasyuk:
For me works fine. Thank you very much.

e-cloud:
@a-tarasyuk , i'm so sorry, in my local dev, i tried to use underscore to do some other thing. And afterward i gave it up. but the code's clean-up wat not thorough.

PR #132, opened 2015-09-28
Avoid labels going outside the element containing the slider

karelcallens:


ValentinH:
Thanks for this PR.

However, you modified the wrong file. Please only modify the src/rzslider.js and then run  ̀grunt` the generate the correct dist files.

PR #134, opened 2015-09-29
labels stay contained within element containing slider

karelcallens:


karelcallens:
Made the changes on the src dir and did the grunt

PR #135, opened 2015-09-29
labels stay contained within element containing slider

karelcallens:
Adaptation to the setLeft for the Low and HighHandle label and the combined label.


karelcallens:
I hope I did right this time. My first PR, so had to find out how to get it working correctly.

ValentinH:
It seems good to me but there is just a little more to be done:
- when using a slider with a single value, it works well on the left (for the min value) but not for the right  (max one). Here's an example: http://jsfiddle.net/xuf9j0ta/

With a range slider, everything seems to be fine: http://jsfiddle.net/ufez818p/

ValentinH:
I think you need to add this to the update lowHandle:
```js
var left = Math.min(Math.max(newOffset - this.minLab.rzsw / 2 + this.handleHalfWidth, 0), this.barWidth - this.ceilLab.rzsw);
this.setLeft(this.minLab, left);
```

ValentinH:
This PR is ready to be merged I think, you just need to rebase it on top of master so there is no conflict ;)

ValentinH:
Since I got no news, I added your code to master manually. Thanks ;)

zebslc:
Brilliant.  You are awesome. :+1:

PR #148, opened 2015-10-11
Seperated to variables.less

emilmoe:
Seperated less configurations and variables.


ValentinH:
Why not importing the variables directly in the rzslider.less file? I don't see why we need an app.less file...

emilmoe:
Of course, that's a solution as well. It's a matter of opinion, if users should be able to import 1 default with configuration or 2.

I have made the changes, for me it's fine either way.

Thanks.

emilmoe:
Thanks. Fast and efficient.

PR #151, opened 2015-10-12
Fixed space between selection bar and handle bubble

emilmoe:
https://github.com/rzajac/angularjs-slider/issues/146


ValentinH:
Good job it does fix it. Could you please run grunt to generate the dist files and rebase your PR so there is only one commit?

emilmoe:
I'm not sure what is going wrong, me and the github app isn't good friends :-) I think I have made the pull request you asked, but something won't merge. Can you check up on that? I will use SourceTree from now as it is more logic to me.

ValentinH:
I just understood the issue: you use the same branch for your 2 PRs: master. To push several PRs on a github repo, it is better to create a branch for each PR. Like this,  you always keep your master up to date with the main repository.

Then if you still have conflicts with the master branch, you can rebase on top of it. Here is a good tutorial: https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request 

So for this PR, I would delete your fork, recreate it, create a new branch, submit a PR and it will be fine... Then once the PR is merged, you can delete the branch and resync your fork with the main repo.

emilmoe:
I understand your point, sometimes I find this over complicated, but I guess it's just me who don't understand git totally yet.

Now I'm not sure actually if you made the changes yourself? I can't make a pull request, but only if you did not change it yet? I don't want to risk making a problem here.

ValentinH:
Indeed, Git and PR/Rebase stuffs are quite complicated at first but they are very powerfull. I had some troubles too at the beginning. ;)

I pushed a change a few minutes ago but there were already conflicts with your PR. The problem is that this PR also contains the edits of your previous one. 

You really should restart you fork from scratch otherwise you will get into troubles again the next time you will submit a PR. :/

src/rzslider.less
```diff
@@ -7,6 +7,8 @@
  * Licensed under the MIT license
  */
 
+@import 'variables.less';
```
ValentinH: See, this update is from your previous PR

emilmoe:
Don't worry, I have deleted it :-) So next time I will be forking a clean copy.

Thanks !

I will try to make another pull request later when I have figured out how to decrease the height of the container when no label nor "translation" is needed.

ValentinH:
OK but you should provide one PR for each fix so the history is better. That fix is cool and I like to see it merged. :)

PR #155, opened 2015-10-16
Update floor label so that it hides correctly when using single slider.

kaspertygesen:
Initial state:
initial state

After change
after change

Before change
before change


kaspertygesen:
```html
<rzslider
 rz-slider-floor="volumeSlider.floor"
 rz-slider-ceil="volumeSlider.ceil"
 rz-slider-model="volumeSlider.min"
 rz-slider-step="50"
 rz-slider-always-show-bar="true"
 rz-slider-translate="volumeSlider.translate"></rzslider>
```
```javascript
    $scope.volumeSlider = {
        min: 4000,
        floor: 50,
        ceil: 4000,
        step: 50,
        translate: function (value) {
            if (value == 50) {
                return "below 50 l";
            }
            if (value == 4000) {
                return $filter('translate')('GLOBAL_ALL');
            }

            return value + " l";
        }
    };
```

ValentinH:
Could you please directly provide a JSFiddle showing the issue?

ValentinH:
Here what I copied from your second message: http://jsfiddle.net/dc089Lnv/
But I don't see the issue.

kaspertygesen:
Ok, it seems like its only a problem when used with the Ionic framework and modal dialogs.
Outside the modal it works just fine as in your jsfiddle.

I have reproduced it here:
http://plnkr.co/edit/WuaObGDkXMitYctKrFUz?p=preview

ValentinH:
OK I see, if it fixes you issue, we can add it. Anyway, I don't understand why it was only calling updateCeilLab()...

Can you run grunt to generate the dist files and add them to the PR? Then I'll merge it.

kaspertygesen:
Ye, that is what it thought as well.
Dist files added.

Thanks for maintaining this great project!

ValentinH:
You're welcome ;)

PR #165, opened 2015-11-02
Forward onStart parameters to callback

sierraindiagolf:


ValentinH:
I just added a commit to pass the sliderId to every callback. Does it correspond to your need? I don't really want to pass more parameters since events you passed in your PR belongs to the private logic of the directive... If one day we want to change this logic, then the onStart will break.

PR #185, opened 2015-11-24
Feature - Add a vertical option to display vertical sliders.

ValentinH:
Simply add `vertical: true` to the options and wrap the slider with a container with a specific height.

Demo: https://jsfiddle.net/dtfbwvh8/


ValentinH:
There is a positioning problem on IE10. Pointers, Ticks and labels are too much on the right.

pihomeserver:
Works in Safari 9.0.2 (latest)

ValentinH:
9bb67f3 fixed the issue on IE10 :)

Issue #157, opened 2015-10-21
Keyboard support


tommhuth:
It would be great if the slider could handle keyboard interaction as well, not to say an accessibility boost. Doesn't need to be any more complex than `tabindex` on `rx-pointer` and arrow left/right increasing/decreasing the slider with the step value set.

Thoughts?

ValentinH:
This is a good idea. How would you handle range sliders?

tommhuth:
Tabbing would select `rx-pointers` which could then be moved with arrow keys. If it's a range slider you would simply tab between the lower and upper `rx-pointers`. Though it looks like the current implementation hides any unused pointer with z-index: that would have to be done with a proper `display: none` or equivalent so that tabbing is not disrupted.

With some added aria roles and tab indices this sounds doable to me. 

One question is how to handle range slider and aria role, as it looks like the current spec does not accommodated for range sliders, and would involve a duplicate `role=slider` attribute on both upper/lower pointers.  Though an `aria-describedby` could be used to differentiate the two.
 
Do you see any difficulties in implementing this?

tommhuth:
Would love to have a crack at this (first fork and merge request it would be), but not sure when I'll get the time for it.

ValentinH:
Actually, I have never used aria roles so I can't tell you what I prefer. ^^
I'm currently working on a 2.0.0 release that will change the way we pass options to the directive so if you start a fork you should wait this release... It should be done by the end of the day.
This release will break current implementations so I need to provide the correct documentation to help users migrating.

tommhuth:
Cool! I'll wait hold the fork, and see what I can cook up.

Looks like mye guess about aria and ranged slider was correct as well: http://oaa-accessibility.org/example/32/

ValentinH:
Any feedbacks on #158 would be great ;)

ValentinH:
Do you know if there is another way than setting a tabindex to enable tabbing to an element? Because the tabindex will need to be customizable since we don't know were the slider can be included... If a user is already using tab index for a form and we set default ones for the slider, it can creates problem.

tommhuth:
That's right, tabindex=0 would let it follow the normal flow. 

The only immediate challenge I see is that each slider dot should have an aria-describedby which would need a unique ID. Any thoughts on how to do that? Also, this describedby text should be customizable so that it can match any language, but I guess a default text value would be acceptable, simply stating “upper/lower”.

tommhuth:
[...]

PR #191, opened 2015-12-05
Add keyboard support [WIP]

ValentinH:
As discussed in #157.
Demo: http://jsfiddle.net/24gryyu5/

Should this be added by default or only if a keyboard option is set?

ARIA reference for sliders: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_slider_role .


ValentinH:
For now, it handles both simple and range sliders. More keys needs to be supported. When a pointer is focused and a user start to update it with the mouse, then keyboard control doesn't work anymore.

ValentinH:
I just added more keys such as page-up/down, home and end. Also, now, when a handle is clicked then it is automatically focused so we can update it after a slide using the keyboard.
I'm not sure of the best way to handle the tracking state of the slider (the`rz-active` class that was applied during slide).

ValentinH:
Focused handle is wrong for range slider when switching min/max with mouse.

georgebatalinski:
Its working great in the fiddle - let me review the commits and see if I have any suggestions. Awesome work

ValentinH:
Thanks! There is at least one bug remaining for range slider (my last comment). Also, it lacks comments and code cleaning.

georgebatalinski:
Got it. I setup the test suite - mocha etc. on your branch [keyboard-support]
https://github.com/georgebatalinski/angularjs-slider.git
Tomorrow I will write some tests, and mock up some examples.

ValentinH:
Great! I wanted to add tests for a long time but never got the time/motivation...

georgebatalinski:
I setup up coverage, as requested :)  [keyboard-support]
https://github.com/georgebatalinski/angularjs-slider.git `angularjs-slider/tests/coverage/`

georgebatalinski:
I am trying to figure out the testing strategy, events are difficult to handle in the unit tests, and aria does not seem to be updating in the tests, we may not be able to rely as heavily on unit tests.

1. Focused handle is wrong for range slider when switching min/max with mouse. - this works fine, I did not see this problem. (+ this is not required, since if the user is going to be using keyboard, they are not going to be using a mouse at the same time) 
2. Is there a reason we are supporting HOME and END - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_slider_role
3. Shift and Shift-Tab is working well on the multiple handles - and that is what we want :) 

Let me know if you have any suggestions on the tests, and if you have anything you would like to test specifically - plus checkout the coverage report.  https://github.com/georgebatalinski/angularjs-slider/blob/keyboard-support/tests/spec/rz-slider-service-test.js

georgebatalinski:
4. May you provide more detail on this one? 
I'm not sure of the best way to handle the tracking state of the slider (therz-active class that was applied during slide).

ValentinH:
1) For me the best behaviour would be that you could start moving the slider with mouse and adjust it a little bit using keyboard. That's why I force the focus (browser state) when a handle is clicked. However, for range sliders, if you switch min/max using the mouse, the inner logic of the library is fine but the focused element (the browser state) is wrong.

On this example, I first switch min/max using keyboard: OK then with mouse: not OK.
msdxyrqnnr

2) I found about HOME and END here: http://oaa-accessibility.org/example/32/ . Also I think they are often use in UI where keyboard are used.

Regarding the tracking state, everything is handled internally using the `this.tracking` property, `rz-active` is only for styling purpose.

ValentinH:
Also, it would be better if you could avoid using JQuery specific function because JQuery is not required for this library. Here's a nice example about using mocked Event without JQuery: http://stackoverflow.com/a/28924708/1713469

georgebatalinski:
Got it. I removed the Jquery and tested using your technique.
https://github.com/georgebatalinski/angularjs-slider.git

1) Nice work on the switching focus with mouse and keyboard :) You are right about producing the same experience for both mouse and keyboard. https://developer.mozilla.org/en-US/docs/Web/Accessibility/Keyboard-navigable_JavaScript_widgets

2) We should add tabindex="0" on the role="slider" - as per  https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_slider_role 
`Flow should be -> Tab to slider first and then we able to tab into the pointer`

3) Quick Feedback 
```
this.maxH[0].focus();
We should make this on function that does the focus()
e.g. 
focusElement(element) {
element.focus()
}
```

```
this.maxH[0]
0 is not descriptive - why is it 0 - is there a significance - what else is in that array? 
```

```
this.tracking
I have no idea what tracking is
should it be model? since it looks like its always - 'rzSliderModel'
```

```
if (this.options.keyboardSupport)
Since we are supporting focus - both keyboard and mouse - wondering if we need that flag. Not sure if it is worth it to disable it, since we have accessibility folks should just have it 
It is also more like FocusSupport
unless there is a reason 

```
Here for the future - if we ever have more attributes 
var attributes = ['aria-valuenow'.....]
forEach(addAttribute)

```

Everything else is awesome 


Let me know  if you want me to look at anything else

**Comment by ValentinH** [comment]
Thanks for this very complete feedback. ;)

About the tabindex="0" on the role="slider", I do it for each handle and I don't agree about the fact that we should first focus the slider and then the pointer. Only editable elements should be focusable so there is no point focusing the whole slider. Quoted from the Mozilla reference: "The slider role is assigned to the "thumb," the control that is adjusted to change the value." Here "thumb" is our pointers.

I agree that we could extract the focus logic to its own function. About the [0] syntax, it's because `focus` is a native DOM method so we need to access the DOM element whereas minH is the JQuery/JqLite object (https://learn.jquery.com/using-jquery-core/faq/how-do-i-pull-a-native-dom-element-from-a-jquery-object/).

For the `this.tracking` property, it can either be 'rzSliderModel' or 'rzSliderHigh'  depending of the moving pointer. (See this line for example: https://github.com/rzajac/angularjs-slider/blob/keyboard-support/dist/rzslider.js#L1205).

For the `if (this.options.keyboardSupport)` condition, my plan is to enable the keyboard support by default to encourage user to provide accessible content but a user should be able to disable it if he really need.

Lastly, about the aria-attributes, I don't understand what you want to do because each attribute gets a different value...

**Comment by georgebatalinski** [comment]
You are fast :) The following are really small improvements.

1. In respect to the ARIA attributes, ngAria does a good job here - https://github.com/angular/angular.js/blob/master/src/ngAria/aria.js
Given that all of them have different values - we could save some lines - like this

```
this.minH.attr({'aria-valuenow': 10, 'aria-valuemin': 20})

```

2. Just for readability then, having something like this - source: Clean Code (book Robert Cecil Martin)
```
var DOM_ELEMENT = 0; (constant or whatever)
this.maxH[DOM_ELEMENT]
``` 

I will create a pull request, for the test suite.

ValentinH:
Damn the list of commits is quite weird now. A PR on a PR was maybe not a good idea! ^^ But I don't know how to do it differently (several people working on the same PR).

I'll try to fix the commits list with a big rebase.

ValentinH:
OK I give up with fixing the commits... Too complicated.

I'll modify my code with your last remarks and we should be good to go.

georgebatalinski:
cool - awesome work :)

ValentinH:
OK do you think we are good to merge?

georgebatalinski:
Yes

PR #201, opened 2015-12-11
Set 'display:inline-block' to the maximum handle.

samechikson:
Previously, the right handle was being set 'display:null' for a range slider, so set it 'display:inline-block' to make it appear.

### DIFF
--- a/dist/rzslider.js
+++ b/dist/rzslider.js
@@ -458,7 +458,7 @@
         if (!this.range)
           this.maxH.css('display', 'none');
         else
-          this.maxH.css('display', null);
+          this.maxH.css('display', 'inline-block');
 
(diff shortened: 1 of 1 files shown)


ValentinH:
Setting display to null is to remove any display value that has been set directly on the element. Thus, the pointer gets its default value which is indeed `inline-block`.

Why did you change that, do you have any problem to use range sliders?

samechikson:
I had to change it because in every instance of the range slider, the pointer was getting hidden. This was the only way to make the pointer show up.

ValentinH:
Can you post an example with JSFiddle please? What version are you using?

ValentinH:
I just received a PR fixing the same issue and I prefer how it's done there: #204 . The problem is due to JQuery I think.

Issue #199, opened 2015-12-10
Is it possible to lock range?


yelnar:
Whenever I change the range its length would stay the same. This would be awesome.

ValentinH:
Is this similar as #196 ?

yelnar:
Yes it is. Now I use `onEnd` option to implement it, but it is little bit dirty so I think it would be awesome to have this feature natively.

ValentinH:
It should not be very hard to implement it. It's the next thing on my todo list ;) 

I'm closing this one as duplicate of #196 .

Issue #196, opened 2015-12-08
How to make handle slider be read only.I am just want  range drag


FantasyLions:
Hi
 In my project  just need drag range.What should I do?

ValentinH:
If I understand correctly what you want, this is not possibly currently but this would be a good feature. If you can describe a little bit more I could implement it quite soon.

FantasyLions:
I am so sorry my English is so poor.I'm trying to say about it
I want to use this slider in the video clip.The video playback time is 50s I just want clip 10s and the 10s is fixed so just nedd drag the rangebar

ValentinH:
Ok so the slider should be draggable like on this  demo http://jsfiddle.net/5mcr5h7r/, but the user should not be able to directly change the min/max values?

FantasyLions:
yes! Is this meaning

ValentinH:
It's available now using the `draggableRangeOnly` option added in 2.3.0.

PR #203, opened 2015-12-16
Add a draggableRangeOnly options. 

ValentinH:
As asked in #196 and #199.
Demo: http://jsfiddle.net/opnrb0hL/

@FantasyLions @yelnar is it what you wanted?


ValentinH:
OK I'll merge #203 before then . I'll do it later today.

FantasyLions:
yes! is awesome.Maby can add this feature:move the whole slider when you drag the handle

yelnar:
Cool! And also I think it is would be better if two edges of slider was draggable too. Thank you.

ValentinH:
Indeed you are both right. I just added what you wanted and also keyboard support.

yelnar:
@ValentinH great, thanks!

PR #204, opened 2015-12-16
fixed reset of maxH element

nibblesnbits:
Calling `this.maxH.css('display', null);` does not reset the style.  Passing `''` does.  This fixes an issue where the max slider bubble does not reappear when updating the model.


ValentinH:
Demo or it didn't happen! :D

ValentinH:
I got it for you: https://jsfiddle.net/7tfbnj92/ (last example)

The problem is when JQuery is used....

ValentinH:
Also, could you generate the dist files by runnning `grunt`please?

nibblesnbits:
updated

ValentinH:
Just to make it perfect: could you rebase on master so there is no conflict, merge everything  so there is only one commit, and remove the temp folder / modified git .ignore .

ValentinH:
I can't merge it as is because there are conflicts.

You need to rebase your branch on master and do the other steps I mentioned above.
Otherwise, I can do it directly but you won't appear as a contributor..

ValentinH:
Merged manually.

Issue #210, opened 2015-12-24
Provide a parameter for the rounding feature


TikiTDO:
Would it be possible to gate this feature: https://github.com/angular-slider/angularjs-slider/blob/master/src/rzslider.js#L638 behind a parameter? 

I have a situation where the slider is used in conjunction with an input field. When a user changes an input field connected to a slider with a sufficiently large step, the rounding function makes it very difficult to enter values.

For example, for a slider with a step of 100,000 it's practically impossible to enter anything without the value getting rounded down to 0.

ValentinH:
OK why not. I'm looking at it right now. Do you want the value displayed on the slider to be rounded or also not to be rounded (when modified from the input)?

TikiTDO:
Actually, sorry. I realized there's a better way of handling this scenario. I just changed my input field to update the angular value on a blur event. Otherwise this will probably become a constant quest to quash all sorts of bugs from this usecase.

ValentinH:
Yes I was also going to tell you that you might use an intermediate value so that the input and the slider are not directly tied.

ValentinH:
Anyway, I pushed a PR so if you really need it, I can merge it. Otherwise, I prefer not to overload the slider with too many specific features.

TikiTDO:
That's probably a good idea. The ceiling and floor made sense from a DRY perspective, since you had to enter that info anyway, but the enforceStep doesn't really make as much sense since you could just not set a step.

PR #211, opened 2015-12-24
feat(enforceStep): Add an enforceStep option (defaults to true)

ValentinH:
Add an enforceStep option (defaults to true) to force the value to be rounded to the step, even when modified from the outside.

As requested in #210

Demo: http://jsfiddle.net/ValentinH/ooakat25/


ValentinH:
Don't want to merge for now. Can reopen if someone really (really) needs it. ;)

PR #225, opened 2016-01-08
Add pointer colour to slider

mtucciar:
Developers are able to dynamically update the pointer colour on the slider.

Demo: http://jsfiddle.net/2n9u63kr/

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,6 +173,7 @@ The default options are:
     ticksValuesTooltip: null,
     vertical: false,
     selectionBarColor: null,
+    pointerColour: 'blue',
     keyboardSupport: true,
     scale: 1,
     enforceRange: false,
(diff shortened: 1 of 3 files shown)


ValentinH:
I added an example to your PR.

I see that you have directly edited the dist file which is not really a good idea, you should only edit the src files and then generate the files by running `npm build`.

Also, is the goal of this new option to change the color dynamically?Somehow like the `getSelectionBarColor` option?

Last but not least, you should also add a test case for this option...

mtucciar:
Ah sorry about that! I was working directly in the dist file because I wanted to explicitly use the included HTML template.

Yes that's correct, but except I left it as a basic attribute instead of creating a function for it. That way when you change pointerColour attribute, the options watch function is fired and updates the CSS accordingly. 

Oh that's right, thank you! Shall I add a test case or will you?

Thanks for the tips :+1: :)

ValentinH:
I just added a guide for contributing: https://github.com/angular-slider/angularjs-slider/blob/master/CONTRIBUTING.md .

Even if it's seems more performant than a function, using a basic attribute will rebuild the slider at each change while using a function will only update the css. So I insist, you should use the same way than I did for `getSelectionBarColor`.

mtucciar:
Thanks for that! :) 

Ahh neat! I learned something new. The next step would be to make all of your suggested changes and then submit a new PR?

ValentinH:
Nope you can just add more commits to this PR. Once eveyrhting will be good, you'll only need to rebase eveything so I can merge and you will be good to go! ;)

mtucciar:
Neat! Thanks a lot for getting back to me so fast :)

ValentinH:
Any update on this?

BlackIPS:
As mentioned at he discussion page for this issue I'd created a pull request by myself.

https://github.com/angular-slider/angularjs-slider/pull/253

Best regards,
Steffen

mtucciar:
Thanks BlackIPS, I haven't had much time out of work to get around to this feature.

ValentinH:
Closing since #253 has been merged

PR #229, opened 2016-01-13
perf(): Remove  inside onChange event

elifer5000:
When the callback run while changing the slider takes a certain amount of time (~50 ms or more), the
 inside onChange can stall resulting in many changes lost along the way while it catches up.
Removing seems to have no visible side effects. It might be possible that this is due to
scope.() being called right before callOnChange(), so the model is updated anyway (see in
applyModel)

See:
http://jsfiddle.net/c8po9y58/3/

Move the slider like crazy and watch the console output, you will see no stalling. To compare use the master version:
http://jsfiddle.net/3bfnocrb/1/


ValentinH:
OK now I understand the issue. :) 
Indeed, in the `applyModel` method we already call `scope.$apply()` so this $timeout is useless and bring  overhead. 

I think I'll also remove the `$timeout` call for the onStart since it should be called right away. However, I'll keep it for the `onEnd` since we need to ensure that the model is synchronized.

src/rzslider.js
```diff
@@ -817,9 +817,7 @@
       callOnChange: function() {
         if (this.options.onChange) {
           var self = this;
```
ValentinH: Just to make it perfect, could you remove this line and call `onChange` directly on this? (When it's done, don't forget to rebase so there is one commit) ;)

elifer5000:
Ok, done!

ValentinH:
Great thanks.

ValentinH:
I just rollback your changes to the callback logic because I found the reason why we needed to wrap them in a $timeout call: if the user was changing one of its scope values in the callback function, then its changed wouldn't be applied. 

In the new version, I replaced the previous $timeout calls by $evalAsync one.

elifer5000:
I see, will the performance be the same, though?

ValentinH:
If I understand correctly how it works, with $evalAsync, if there is already a digest in progress then it will be executed in it, otherwise it will program a new one.

elifer5000:
Indeed it seems to be ok. I tried again with my same sample and there's no
dropped events. You can try it here:
http://jsfiddle.net/3bfnocrb/2/

Issue #243, opened 2016-02-02
Show selection bar from 0, when negative


levipadre:
Hi,

Thanks for the directive, it's pretty cool.
I was wondering, that how can I set up the selection bar from zero.
[http://jsfiddle.net/7b15e7wt/](http://jsfiddle.net/7b15e7wt/)

Floor value is -10, ceil value is 10, init value is 0. But unfortunately the line starts from -10 and not 0.

Is that possible somehow?

Thanks

ValentinH:
Currently you can only show the bar from the left (`showSelectionBar`)  or from the right (`showSelectionBarEnd`). 

Just to be sure of what you need, tell me if I get it correctly:
- if model = 5 : -10 ---------------------------------|==========O------------------- 10
- if model = -5: -10 ---------------O=========|--------------------------------------- 10

levipadre:
Yes, this is exactly what I need.

ValentinH:
OK do you think you can try to do it in a PR?

levipadre:
Perfect, it's working like a charm! Thank you so much, it was so quick!
I will watch when you push to master, until then I'm using this branch.

Cheers!

ValentinH:
I just need to add the test coverage so it should be done by the end of the week.

levipadre:
Sounds perfect, thanks!

ValentinH:
PR merged in master.

PR #250, opened 2016-02-03
feat(selectionBar): Add a showSelectionBarFromValue

### LINKED ISSUE #249: showSelectionBarEnd broken with showTicks=true
The ticks are displayed as if the option was showSelectionBar (on the left of the value).

ValentinH:
Add a showSelectionBarFromValue to display the selection bar from a specific value as requested in #243. It also fixes #249 .

Demo:  http://jsfiddle.net/L96ensbw/


ValentinH:
CC @levipadre

Issue #259, opened 2016-02-11
Ability to specify steps, but not be fixed to those steps


lookfirst:
I've got a slider of 1-200. 

I'd like to show ticks in steps of 10, but be able to drag to numbers between those steps.

ValentinH:
This is not currently possible but I have already thought about it... I need to find the best way to provide such a feature.

lookfirst:
:+1: @ValentinH

ValentinH:
I have started some work on #264

Paul-Todd:
Love the implementation in #264 exactly what I was looking for!

ValentinH:
#264 merged and available in 2.12.0 release :)

lookfirst:
Thanks @ValentinH 👍 ❤️  :metal:

PR #264, opened 2016-02-13
feat(ticks): Accept numbers for showTicks/showTicksValues

ValentinH:
Thanks to this feature, you can display ticks at intermediate positions without setting a fixed step.
As requested in #259.
Demo: http://jsfiddle.net/xaa0qox1/

A few problems found so far:
- [x] clicking on a tick should set the slider value exactly to the tick value 
- [x] clicking on a tick value should set the slider value exactly to the tick value
- [x] when using `showTicksValues`, the pointer label is hidden and so we don't really know the current value. This is because when the `showTicks*` options were only boolean, the labels were useless since they were replaced by the ticks values.
- [x] complete unit tests


lookfirst:
Nice work!

The UX that I've seen used elsewhere is a bit different. The tick value is below and the slider tick value is shown above as it is dragged. It hides when the mouse is not over it.

http://take.ms/aenJI  - with the mouse over the slider (shows the value above)
http://take.ms/zbbhu - hidden when the mouse is not over the slider

I can also click on the values below the line to set the slider value.

ValentinH:
In my opinion, hiding the value when the mouse is not on the slider is not an option since it is important for the slider to be fully understandable without any additional widgets.

Thanks to CSS, ticks can already been set on the bottom but the slider need to work properly for ticks with top and bottom.

lookfirst:
Sure, I agree with you - I don't like hidden UX.

ValentinH:
@lookfirst  I have updated the demo and add handling for showTicksValues. Can I get your feedback please?

lookfirst:
:+1:

ValentinH:
Ok then I just need to complete to unit tests.

johnsheldon:
Any ETA for the release which contains this feature? We just so happened to need it and saw it was being implemented.

ValentinH:
Normally, the feature is ready, only the unit tests are missing. I just need to find the time to write them...

Does the demo looks good to you?

lookfirst:
Just integrated this into my app... thanks! screenshot: http://take.ms/RIYCH

ValentinH:
Great! :)

PR #266, opened 2016-02-16
feat(rzSliderOptions): Change rzSliderOptions to use expression binding

### LINKED ISSUE #265: Use expression binding for rz-slider-options
The current code uses two-way binding for the slider options, which can cause infinite digest loops (see https://jsfiddle.net/txupy024/). The slider code never modifies the options value, so the two way binding is not needed.

The fix is a two line change, and I would be able to make the contribution myself.

rabbitfang:
Switch from using two-way binding for rzSliderOptions to using expression binding. Using expression
binding enables more complex use cases such as using the return value of a function as an options
object.

Closes #265.


rabbitfang:
If more detailed tests are needed, I can add them. I wasn't quite sure what I should test. I added the `optionsExpression` option for the test helper for setting the expression that is used for directive (defaults to `slider.options`).

rabbitfang:
Example JSfiddle for why this is useful: http://jsfiddle.net/fvb1uw23/

ValentinH:
I have to admit that at first I didn't really like it and it feels weird to me, but it works without creating breaking changes (as far as I have tested for now).
I'm just curious of where have you seen such a way to pass parameters to a directive? I was thinking that expression bindings where only good for calling callbacks from a directive...

rabbitfang:
I personally use expression binding when there is no need for the directive to make modifications to a value (and where attribute binding wouldn't make sense). Expression binding allows the greatest freedom for the consuming developer. One common example for when AngularJS itself uses expression binding over two-way binding is the `ngShow`, `ngHide` and `ngIf` directives. By using expression binding there, I can use something such as `ng-if="!shouldHide"`. It also allows filters to be in the parameter.

For rzslider, I use the return value of a function that dynamically generates the slider's configuration based on various parameters that it pulls from various locations. Even though the `$watch` uses `angular.equals` comparison to check for updates, the two way binding only checks for identity. If the function returns a new object on every call, a infinite digest loop is created.

ValentinH:
Well, looks good to me! Thanks :)

Issue #270, opened 2016-02-18
RTL Support


Liam-Ryan:
Is it possible to switch the direction of the slider by setting a flag for right to left? I haven't been able to find anything in the documentation and switching floor and ceiling seems to work as far as calculating values but the sliders stay stuck at the rightmost extreme of the bar

ValentinH:
Nope this is not currently possible but this could be a cool feature.

For now, you can do a work-around by using the `stepsArray` and `showSelectionBarEnd` like in this example: http://jsfiddle.net/hm9nntmm/

Liam-Ryan:
Thanks I'll see if I can work it in easily at this stage. Would you mind
leaving this open and I'll see if I can put together a pull request in the
near future?

ValentinH:
Yes sure, I'd be glad to provide support for it if needed ;)

Liam-Ryan:
I'm just starting to work on this feature, what approach would you recommend for quick testing and deployment? I notice grunt serve does not run the demo, is there a task that watches files and runs the demo?

ValentinH:
`grunt serve` should serve the demo at `localhost:9000/demo/index.html`.
Then to test your new feature, simply add a block in the demo.html and demo.js.

Liam-Ryan:
Thanks I'm making good progress but the bars and labels are aligned in the incorrect direction when I enable RTL, could you point me towards the logic which controls drawing the bars or even just the variable that is used to refer to the range divider bar rather than the overall slider bar? 

![image](https://cloud.githubusercontent.com/assets/6696533/13199061/6c8c1b60-d812-11e5-99ac-cfc51e5da615.png)

Liam-Ryan:
Hi Valentin, 

I'm almost ready to commit a PR but I've noticed some differences in the code style in places

` if (this.tracking === 'rzSliderModel' && newValue > this.scope.rzSliderHigh) {
                if (this.options.noSwitching && this.scope.rzSliderHigh !== this.minValue) {
                  newValue = this.applyMinRange(this.scope.rzSliderHigh);
                }
                else {`

vs 

`  if (newOffset <= this.dragging.lowLimit) {
              if (this.minH.rzsp === 0)
                return;
              newMinValue = this.minValue;
              newMaxValue = this.minValue + this.dragging.difference;
            } else if (newOffset >= this.maxPos - this.dragging.highLimit) {
              if (this.maxH.rzsp === this.maxPos)`

and 

[...]

PR #276, opened 2016-02-21
feat(RzSlider): Add RTL support

Liam-Ryan:
Added rightToLeft option which will reflow slider from right to left ( or switch vertical if vertical: true)
#270 RTL Support


ValentinH:
Could you set up a JSFiddle showing this new feature in practice? :)

Liam-Ryan:
This one is more for feedback on the code, I still have to work out the range slider movement. Also I'm working out how to git as I go so let me know if I need to start again and do it differently :)

ValentinH:
OK i'm taking a look at your code then.

Liam-Ryan:
Thanks, also I just realised that the rightToLeft default is set to true currently so that's why so many tests fail. You can delete this if you like and I'll open a fresh pull request when I have sorted that and the range dragging. To be honest I didn't expect the PR to go straight to the main branch instead of my fork :)

ValentinH:
I don't need to close it, you can just correct whatever you want on your branch and force-push it (git push -f).

ValentinH:
Besides, I think that there is a lot of modifications for only one feature and I don't really like this. I don't really want to see `if(this.options.rightToLeft)` on every function of the library so we need to simplify the way to do it.
However, I don't have currently any better suggestion to provide but I'll try to think about it.

ValentinH:
I just checkout your code and for what I've tested so far it's working well! :)

Liam-Ryan:
Thanks, I have noticed I missed ceiling labels and I think the range slider will be a bit tricky so I don't know if I'll get everything sorted out today ( I need to do something besides code on my weekend :) )

For now I will continue to try and get core logic working correctly and once the logic is correct and all tests written I will try to modularize the feature in a better way

ValentinH:
Sounds good :+1:  (And yes don't forget to enjoy the week-end :p)

Liam-Ryan:
Just some minor bugs on the floor label positioning left to fix then update unit tests ( assuming I don't find more major issues when writing the rest of the tests ;) )

I've committed the latest to give a better idea of the changes to the code and to help figure out how best to integrate. We could create calls to helper functions wherever we're checking if rightToLeft is true or not to tidy up the code a bit.

I'll finish up the minor bugs and unit tests then wait to hear back from you on what you'd like me to look into (is there a better forum than the PR to discuss things like this?)

ValentinH:
I think that the PR is OK to discuss about it since everybody can see and discuss about it.

Liam-Ryan:
Hi Valentin, 

When you have a minute could you check the three tests that are failing? If you could explain the test cases to me it would be helpful. I debugged in chrome but from what I can see the expected results do not make sense, unless I am missing a feature. 

Could you explain what these tests are checking and pause them in a browser to see if the RTL behaviour is as expected or if something is missing? 

Thanks again

ValentinH:
For the first test, the problem is that I used the `valueToOffset` function to compute the move position but since we are using RTL, the position is not the one we expected at first: if you want to move by the offset of the value 10 then you need to pass `max-10` to `valueToOffset`.

For the second and the third tests, to move left, you need to use negative values. Indeed, with RTL `maxPos` corresponds to the rightmost offset while the max value is on the left of the slider.

Liam-Ryan:
Thanks I'll need to check again I thought i had already altered the offsets. I'll try to finish out the tests tomorrow

Liam-Ryan:
I believe the unit tests and core logic are now correct and complete, I went overboard with the unit tests and duplicated most of the existing ones while setting rightToLeft to true to allow for more freedom when refactoring the feature. If there are too many then we can remove them before the merge if you like.

Have you any suggestions on how you would like the feature structured? I won't be starting until tomorrow but I think right now I would probably lean towards having helper functions where there are currently switches and reassigning the helper functions depending on the rightToLeft option. Eg
`          if (!this.options.rightToLeft) {
            this.scope.ticks.push(tick);
          } else {
            this.scope.ticks.unshift(tick);
          }`
becomes
`pushTick(tick)`

pushTick is assigned based on whether the option for rightToLeft is enabled or not. What do you think?

Liam-Ryan:
Do you think the two unit tests that are failing are legitimate failures? They work fine in Chrome but PhantomJS seems to set the precision really high on the css property.

ValentinH:
For the two failing tests, yes this is quite annoying.  I was going to recommend using the `closeTo` assertion but we are handling strings with 'px'.

The easiest way I found is to use getBoundingClientRect() and floor the expectedValue:
```js
var expectedDimension = Math.floor(helper.slider.valueToOffset(8) + helper.slider.handleHalfDim);
expect(helper.slider.selBar[0].getBoundingClientRect().width).to.equal(expectedDimension);
```

For the refactoring, here's a few thoughts:
- avoid using too long ternary expressions like [this one](https://github.com/Liam-Ryan/angularjs-slider/blob/pr/rtl/src/rzslider.js#L1233), if/else are easier to read. ([this other one](https://github.com/Liam-Ryan/angularjs-slider/blob/pr/rtl/src/rzslider.js#L1461) is OK IMO).
- for duplicated blocks like [this one](https://github.com/Liam-Ryan/angularjs-slider/blob/pr/rtl/src/rzslider.js#L1602) or long if switches, I think it would be cleaner to create temp variables to handle `minH, maxH, minValue, maxValue`, initialize them based on `rightToLeft` and then having only one logic block. BTW, in the linked line, I think that `maxH.rzsp` should be `minH.rzsp`.
- the current `shFloorCeil`, is a mix of the 2 previous points and can also probably be simplified by using temp variables for `minLab, maxLab, flrLab, ceilLab`.
- the actions in `onKeyboardEvent` are not really readable for now. Maybe a simple, `getKeyboardActions()` could be more readable with 2 different dictionaries based on `rightToLeft` .

The other modifications are fine to me and don't need refactoring. For instance, the ticks handling part is already readable to me.

Last but not least, don't forget the semi-colons (just to be consistent with the existing code) and reformating the code using the editor config.

Liam-Ryan:
I tried using the editor config but I must have mine set up differently, it
threw out all of the indents for the entire source code and I needed to
re-merge with the base version. Could you let me know what your webstorm
tab settings are? Or is there a file to set these in the project that I
missed?

ValentinH:
I think it is the 'Indent chained methods' that needs to be off in the Code style>JS settings.

I had to do this to keep the same style as previously (when I was using Atom).
image

Liam-Ryan:
Unfortunately even with these settings my IDE does not match, I got close by setting all numbers in the screenshot to 2, not 4. There are still differences in function declaration style when I compare with master after ctrl + alt + l and it doesn't seem to alter my entered code. Am I missing a JSHint or code style template in the project? 

On a separate note I've tried something in the latest commit which I think is quite neat but I can remove if you hate ( getHandleLabelPos)

ValentinH:
Are you sure that the editorconfig is used? Indeed, my default settings are tab with 4 spaces but for this project, the editorconfig specifies 2 spaces:
image
Also, I can see that in your latest commit, all the function declarations have been changed by adding a space after the "function" keyword. For this, my settings are:
image


Finally, I don't see the `getHandleLabelPos` in your last commit.

Liam-Ryan:
Sorry I must have done something silly, I'll send it up again after work
today

Liam-Ryan:
I've pushed the missing commit for the getHandleLabelPos. I don't think I will get a chance to do much tonight, I will aim to complete the PR by Saturday

ValentinH:
The `getHandleLabelPos` helper is cool: :+1: ! Since the min/max logic in [this line](https://github.com/Liam-Ryan/angularjs-slider/blob/35611db1cab8a702bcdea1a3656fff94941be222/src/rzslider.js#L910) is smaller, it can be set on a single line I think.

A little remark, it's better if you use single quote to remain coherent with the code base. ;)

Liam-Ryan:
Hi Valentin, 

I have finished a run-through on the refactoring. Some things are still not as nice as I would like but I've been staring at it for so long I think I need fresh eyes to see what else might be improved. 

Let me know if you would like anything changed and if my editor settings are ( finally ) matching master. 

Btw I have already merged the roundstep functionality from master, not sure if there is a unit test added for this but I believe rounding both values should be fine in each instance.

ValentinH:
It looks pretty good to me.

My last point would be to simplify all the duplicated tests by keeping only the ones needed to get a 100% coverage. Thus, future maintenance will be easier.

Liam-Ryan:
No issue deleting the tests, although they could be useful in future in
case there are unexpected side-effects if changes aren't tested in RTL.

For instance I was going to delete all tests but those testing right and
left in draggableRangeOnly-range-slider-test.js but this could result in a
future change altering the behaviour of page up and page down when RTL is
enabled being missed by testing. Similarly the other tests are there as a
safeguard that the behaviour in RTL is as expected since I expect a lot of
pull requests may forget to test the behavoiur in RTL.

Let me know whether I should go ahead and delete the duplicates anyway.

ValentinH:
Just to be sure: I don't want to remove all duplicated tests. All function that have specific conds using RTL should be tested with RTL=true and rtl=false. I didn't read all the tested you have added but if some of them don't test RTL specific cases they might me removed.

Liam-Ryan:
What I meant by my ramble was should I take out tests that are not actually different in RTL but test the expected behaviour. 

The example I gave was that page up, page down, home and end have the same behaviour in RTL as LTR. 

I tested them for refactoring but they may be useful for future instances where somebody may change the keyboard actions code and accidentally swap page up and page down in RTL.

If you don't think these sorts of tests are useful I'll take them out and just keep the tests where behaviour is changed ( keyboard right, keyboard left ).

ValentinH:
Indeed, keyboard behavior is OK to be tested twice, you are right.

I was more referring to tests about functions that don't use rtl at all, like these ones: https://github.com/Liam-Ryan/angularjs-slider/blob/pr/rtl/tests/specs/accessibility-test.js#L226

Liam-Ryan:
Ok that makes sense, I'll start looking through them now

Liam-Ryan:
I think that's it, unless I misunderstood the criteria for removing tests

ValentinH:
I think that there are still useless cases. For instance, for the ticks, only the `showTicksValues: true` should be tested with rtl since other cases are not specific to it.

Finally to complete the PR, you will need to rebase/squash everything to get only the first commit message, add a block in the demo page, update the ReadMe and the Changelog.

Liam-Ryan:
Do you want a separate block showing everything in RTL or will I just go for one example as well as the rightToLeft option on the bottom slider?

ValentinH:
I think that one block is enough like for the "Slider with custom floor/ceil/step" for instance.

Liam-Ryan:
I rebased but got plenty of conflicts while doing so, is that normal? Is there some way I can check everything is ok before I do a git push -f ?

ValentinH:
It seems to be fine since there is only one commit and the same number of modified files as previously.

Just in case, can you check with your local copy that files are the same?

Liam-Ryan:
It looks like it didn't copy properly, the copy I have is before re-factoring. I have run the demo with global rtl of true which seems fine and all unit tests are passing.

ValentinH:
I was going to merge but I have a question: is it normal that vertical sliders are also switched (max on bottom, min on top)?

Liam-Ryan:
I'm not sure if it's normal in RTL language but I can see why someone may
have a use-case for it and we were getting it for free anyway. I've noted
in the readme.md that vertical:true and rightToLeft: true will reverse the
vertical sliders and mean that left increments and right decrements them.

ValentinH:
Yes, I was reasoning wrong: "what if I want to use RTL and vertical from top to bottom" but I simply need to set vertical: true and rtl: false.

Let's merge!

Thanks for all this great work! :+1:

Liam-Ryan:
No problem, thanks for all your help through the process

ValentinH:
Red Label! The floor ceil is not shown anymore! I need to find out why...

ValentinH:
The (not) funny part is that it's the only part that is not directly tested! ^^

Liam-Ryan:
Of course ^.^

I think this is a case of mis-merging, I do remember I had this issue early
on in development. It looks to me like it's linked to flrLab.rzsp. I'll see
what I can find

[image: Inline image 1]

Liam-Ryan:
I think it's the updateFloorLab function, testing now

      updateFloorLab: function() {
        this.translateFn(this.minValue, this.flrLab, 'floor');
     *   var position = this.options.rightToLeft ? 0 : this.barDimension -
this.flrLab.rzsd;*
        this.setPosition(this.flrLab, position);
        this.getDimension(this.flrLab);

ValentinH:
Indeed, the condition is inversed.

ValentinH:
I fixed it. I'm double checking to see if everything is OK and I bump the release.

Liam-Ryan:
Ok, sorry about that :$

 I had just noticed the coverage missed 4 lines in the shFloorCeil between
my PR and the merge to master and was checking that when you found this. I
think they are linked maybe?

Liam-Ryan:
Oh I just realised - it probably didn't run coverage after I rebased and
used git push -f, would that be correct?

ValentinH:
No this part has never been covered actually. I just added the tests that would have catch it so we won't miss it next time. There are still tests missing for the positioning of all labels but they are quite hard to test since they depends on the labels width.

Liam-Ryan:
Ok, I'm finishing up for the night but if you need me to do anything just
post here and I'll check it first thing tomorrow

ValentinH:
No I'm also done for tonight, thanks.

Actually, you were right, the coverage has dropped to 99% and it was due to this bug: I don't understand why codecov didn't show any warning for it. Probably due to the several rebase that lost it...

PR #279, opened 2016-02-24
Fixed number precision when draggableRange=true

radomik:
When there was set options.precision>0 (non-zero fraction part) and moved a range there was a issue with number rounding on both ends of selected range. Sometimes numbers had small value at far fraction.

It can be reproduced with live demo by setting, for example:
  minValue: 3.6,
  maxValue: 5.8
  floor: 0,
  ceil: 10
  step: 0.1,
  precision: 1,
  draggableRange: true

It's possible to produce value, e.g: 6.2999999999... but used precision expects to see rounded value: 6.3.

Before: http://jsfiddle.net/egbvgq1n/
After: http://jsfiddle.net/737jbz10/


ValentinH:
Could please provide a before and an after demo to understand what you fixed?

ValentinH:
I have created the 2 demos manually but you forgot to generate the dist files so the "after" demo doesn't use your updated code.

Please follow the CONTRIBUTING.md file.

radomik:
Thank you for reply. I have read now CONTRIBUTING.md file now. I will do everything as described there and create new pull request later today.

ValentinH:
You actualy don't need to recreate a pull request. You can simply rebase all your commits and force push them into this one.

radomik:
I have it already done. Is everything ok?

ValentinH:
Could you rebase/squash everything so there is only one commit?

radomik:
Now should be OK. Sorry for troubles, I'm still learning GIT.

ValentinH:
Thnaks :+1:

PR #280, opened 2016-02-25
feat(logScale): Implement a logScale option

ValentinH:
Implement a logScale option to display sliders with logarithmic scale.
Demo: http://jsfiddle.net/v6spvvpg/

Todo:
- [x] handle ticks


BrunoCartier:
This would be a useful feature :+1:

ValentinH:
I could merge it but it won't support ticks cause I couldn't find a way to display them correctly.

seb2578:
Hi Valentin,
Greetings from a French expatriate from NYC.
Would you consider merging this pull request (even without ticks) ? I would definitely use it if it was made available.
Sébastien

ValentinH:
OK I'll merge it. Just need to add tests then. ;)

ValentinH:
This was merged and released under 5.7.0

PR #289, opened 2016-03-09
Allow to show less ticks than steps and allow to hide pointer labels

fofi:
This allow to add step for ticks in case you want less than steps.
It also add an attribute to hide pointer labels, before always hide pointer labels when tick labels were shown.


ValentinH:
Could you please set up a JSFiddle showing what feature you have added?

fofi:
Here it is:  http://jsfiddle.net/fofi/7au7vsL6/1/

src/rzslider.js
```diff
@@ -595,14 +602,12 @@ function throttle(func, wait, options) {
         this.hideEl(this.ceilLab);
       }
 
-      if(this.showTicksValue) {
```
ValentinH: Why are you removing this part? This will break existing implementations...

ValentinH:
The `rz-slider-hide-pointer-labels` doesn't handle true/false values. If you set false, then the pointer labels are hidden which is not what I would expect.

Btw, I have started some work on the latest version to add the same feature: #264

fofi:
Hi, 
that part of code, auto hide all labels when tick labels were active, so I added the `rz-slider-hide-pointer-labels` to control it.

About the true/false stuff, at least in this version, that occurs to every attribute, and I was not going to fix all that for now, I just add it like it was.
The directive get it 'true' as string and this make `value = !!"false"`, so that ends in true because the string exist.

This probably is due to the directive version we are using, and we will try to upgrade as soon as possible.

ValentinH:
Do you use NPM to grab this library code?

fofi:
Not at the moment, so i think its okey.

ValentinH:
If you don't use it, I would prefer that you use your patched version directly and not update the legacy 1.1.X version. I only keep it in case of a major bug being discovered but I don't want to add new feature to it.

Issue #272, opened 2016-02-19
Prefix all classes with rz-


foucdeg:
### Steps to reproduce
1. Create a slider with discrete steps : 
```javascript
var slider = {
      min: 3,
      max: 6,
      options: {
        floor: 0,
        ceil: 9,
        stepsArray: 'MLKJIHGFED'.split(''),
        showTicksValues: true
   }
};
```

### Expected behaviour
The slider elements should have class all prefixed with `rz-`, such as `rz-bubble`

### Actual behaviour
Individual ticks have a class called `tick`. 

Tick is a common name that could be used in other parts of a website so it's probably not best to use such a generic classname inside a component. In my case I was using a `tick` class for a form which shows a tick when a field is valid, I added CSS to that class which impacted my sliders.

ValentinH:
:+1:

ValentinH:
Fixed and release in 2.10.3.

PR #292, opened 2016-03-13
Prefix all classes with rz- #272

funny-vince:
Prefix all classes with rz- to avoid conflicts with other CSS classes (e.g. from an other library) [#272](https://github.com/angular-slider/angularjs-slider/issues/272)

I was trying to fix a bug on a friend's project and I realized that another `vertical` class was overriding the left property of the angularjs-slider `vertical` class. I came here to report the issue, but it was already reported so I decided to fix it. I prefixed `tick`, `vertical` and `selected` with rz- (now `rz-tick`, `rz-vertical` and `rz-selected`) to avoid further conflicts with those class names.


dist/rzslider.js
```diff
@@ -764,7 +764,7 @@
             tick.tooltipPlacement = this.options.vertical ? 'right' : 'top';
           }
           if (this.options.showTicksValues) {
-            tick.value = this.getDisplayValue(value, 'tick-value');
+            tick.value = this.getDisplayValue(value, 'rz-tick-value');
```
ValentinH: This line should not be modified in my opinion. It's not about css class but it is a simple key that is used for the translate option to determine which value we are translating.

ValentinH:
Except the comment above, this is a really good job. :)

funny-vince:
I updated my pull request. Thanks for your feedback :)

ValentinH:
Great job! :+1:

PR #299, opened 2016-03-22
Allow custom translate function when slider steps are specified

nparrish:
Currently, if the user specifies a steps array, their translate function is ignored.  This changes that.  When a translate function is specified, it will be passed the step value (the array index), the slider ID, and the slider label.


dist/rzslider.js
```diff
@@ -423,9 +423,16 @@
           this.options.floor = 0;
           this.options.ceil = this.options.stepsArray.length - 1;
           this.options.step = 1;
-          this.customTrFn = function(value) {
-            return this.options.stepsArray[value];
-          };
+          if (this.options.translate) {
+            this.customTrFn = function(value, sliderId, label) {
```
ValentinH: Why don't you just do:
```js
this.customTrFn = this.options.translate;
````
?

ValentinH:
Why do you need to use a translate function while using stepsArray?

nparrish:
For each point on the slider, the displayed text should be different depending on whether it is the high end or the low end of the slider.

dist/rzslider.js
nparrish: Good point.

ValentinH:
Thanks for the updates. Could you add a test to complete the coverage? Also, before I can merge, you need to rebase and squash all your commit into a single one.

nparrish:
Done.  Thanks for the feedback, and let me know if you'd like to see any other changes.

ValentinH:
One last thing: can you rebase and squash everything so there is only one commit?

nparrish:
Commits squashed!

ValentinH:
Thanks ;)

PR #306, opened 2016-04-01
fix(rz-vertical rz-tick-value): Position long labels on vertical sliders correctly

### LINKED ISSUE #305: Long labels wrongly positioned for vertical sliders
### Steps to reproduce
1. Use quite long labels
2. Set orientation to vertical

Demo: http://jsfiddle.net/nqnwt57n/

### Expected behaviour
Labels should be correctly positioned at the left of the slider

### Actual behaviour
Labels starts from the right and are over the slider

### Fix
Use left position value instead of right.

dbartholomae:
Fixes #305 

I manually checked whether this works by using the same change on http://angular-slider.github.io/angularjs-slider/
But given that I don't now the code base that well and style changes are not covered by tests, you might want to check this again.


ValentinH:
30px looks to far to me, I would use `left: 24px`. What do you think?

Here's a demo with you code: http://jsfiddle.net/y57c3bco/

dbartholomae:
Personally I would prefer a little more spacing (`left: 30px`). For backwards compatibility you might want to choose `22px` though, as this seems to correspond to the old positioning :)

ValentinH:
Indeed, the goal is to keep the same style as before and to have the same space for horizontal and vertical sliders.

Then you can still override the value in your custom CSS ;)

dbartholomae:
Forcepushed the change to `22px` to this branch :)

ValentinH:
Great thanks ;)

PR #323, opened 2016-05-12
Add BoundPointerLabels Option

andreytkachenko:
Added new boolean option - boundPointerLabels (default: true).
On True: nothing changed.
On False: pointer label will not stick to the ends of slider.


ValentinH:
Could you add a demo please by forking the default jsfiddle?  :)

andreytkachenko:
Hi, here is the demo https://jsfiddle.net/andreytkachenko64/2v6173zs/5/

The idea is to allow pointer labels move like it was in 1.x versions

ValentinH:
It only works for the value label but don't you think it should have the same behaviour for floor and ceiling labels?

Also unit tests covering your changes are missing.

andreytkachenko:
Only pointer labels are movable, so it is impossible case for floor and ceiling labels because they are fixed in one place (and it is easy to locate them through css). But I didn't find any good way to move this logic to css (instead js parameters).

ValentinH:
Thanks ! (and sorry for the delay of the merge)

PR #325, opened 2016-05-18
Adding SASS files

aborchew:
Not sure if you even want this, but I took the time to do it, so I figured I would share.


ValentinH:
Thanks for taking the time to submit this PR. However, I don't want to merge it since this project is using LESS and I don't want having to maintain 2 codebases. If one day someones find an issue with LESS that can be solved by switching to SASS then I will switch to it.

Issue #322, opened 2016-05-12
maxRange option?


envision:
It seems you have not implemented maxRange option...is there a way to achieve this by hooking into some event or so?

ValentinH:
Indeed this isn't implemented but could be added quite easily, just need to mimic to minRange behaviour. If you want you can submit a PR and I can guide you if needed.

envision:
@ValentinH I tried to create a fork for this but I haven't ever written tests for front-end code so this is too complex project for me to hop in out of the blue...

ValentinH:
I think there are only a few tests to be added. Just search for 'minRange' in the specs folder, copy/paste those tests and adapt them ;) You can already create a Pull request and I can guide you there.

envision:
How can I test my changes with a browser without building dist files in each turn?

ValentinH:
Sorry, I've missed your last answer. I just finished implementing the option...

While developing, you should run grunt watch and everything will be generated automatically.

ValentinH:
Implemented in #333 and released under 2.14.0

PR #333, opened 2016-05-22
feat(maxRange): Add a maxRange option

ValentinH:
As requested in #322.

Demo with maxRange=150: http://jsfiddle.net/hhfb040m/


envision:
It's working, thanks! :)

Issue #328, opened 2016-05-19
Slider value does not match custom values specified


HemanthD:
### Steps to reproduce
1. Include a slider with customized stepsArray with values and legends
2. Have the values different from the 'index' value (for e.g., in the demo, the last value (index: 4) is set to 5
3. Check the slider value - the value matches the 'index' and not the 'customized value'
4. The pointer label shows the correct custom value

Demo: https://jsfiddle.net/HemanthD/vqwa3m0x/2/

    vm.customSlider = {
	        value: 0,
	        options: {
	            ceil: 5,
	            hideLimitLabels: true,
	            hidePointerLabels: false,
	            showTicks: true,
	            showTicksValues: false,
	            stepsArray: [{value: 0, legend:'Not Applicable'}, {value: 1, legend:'>= 35%'}, {value: 2, legend:'25% to 35%'}, {value: 3, legend:'15% to 25%'}, {value: 5, legend:'< 15%'}],
	        }
	    };

### Expected behaviour
Slider value should track the customized value specified in the stepsArray. The value associated with the last tick should be 5.

### Actual behaviour
Slider value should tracks the array index and ignores the customized value specified in the stepsArray. The value associated with the last tick is shown as 4.

ValentinH:
First when using stepsArray, you should not set te ceil property .

Then the value of the slider when used stepsArray is by design the index of the array. Indeed, the value can be anything and not only numbers.

HemanthD:
Thanks for the clarification. The use of ciel does not change the behavior.

I believe there should be a way to access the selected value as specified by the stepsArray (for e.g. an attribute like stepValue) as this would be useful while using stepsArray.

ValentinH:
Yes, I might consider setting the `rz-slider-model` and  `rz-slider-high` to the selected value when used with stepsArray. It would be a breaking change and so it would be done as a major version update, 3.0.

ValentinH:
From 4.0, rzSliderModel and rzSliderHigh are bound to the actual value. ;)

ValentinH:
Your demo is not working as expected: https://jsfiddle.net/1cdw90pu/

HemanthD:
Thanks!

PR #335, opened 2016-05-25
refactor(stepsArray): Bind rzSliderModel and rzSliderHigh to the actual value

ValentinH:
As discussed in #328.

Demo: http://jsfiddle.net/55wh6ybw/


ValentinH:
I think I am done with the refactoring, but I need to fix a lot of tests... \o/

PR #357, opened 2016-06-29
Hide floor/ceil label when overlapped on combo label

### LINKED ISSUE #356: combo label piles up on floor/ceil label
Get's obvious when setting long label with `translate` option.

pileup

[fiddle](https://jsfiddle.net/cxqq5aa0/4/)

Liooo:
Fixes #356 

[Working fiddle](https://jsfiddle.net/cxqq5aa0/5/)


ValentinH:
Thanks for taking care of that!
I think you can remove the cmbLabelShown property  because it's never read.
Also, it will be good to add a test case covering this bug. If you want, I can point you the place the do it ;)

ValentinH:
For the unit test, you can add a case in this file: https://github.com/angular-slider/angularjs-slider/blob/master/tests/specs/labels-test.js

Liooo:
cmbLabelShown is actually used [here](https://github.com/angular-slider/angularjs-slider/pull/357/commits/24e3dbbe3cdb3079411b70c3f741fd49d282daac#diff-06d2fc419ad22007df6fe972317d04cbR1110) ;)

in order to test this case accurately, we need to take care of viewport size actually I think. The ideal test scenario is to ensure the cmb label is hidden when
- max/min label element is hidden and combo label is shown
- max/min label element (which is invisible) IS NOT overlapping the floor/ceil label
- cmb label element IS overlapping the floor/ceil label
, which is not that easy to setup.

The added test sets `translation` that returns loong string and expect cmb label to be hidden. But the same test also passes in the previous library version, because min/max label elements overlap floor/ceil labels either way.

(same thing applies to any label tests where the handle is located at middle, precisely speaking)

ValentinH:
My bad, I was on mobile this morning and miss it and also the unit test you added :)

ValentinH:
LGTM 👍

PR #358, opened 2016-06-29
Expose label on scope in template

### LINKED ISSUE #355: Changing the way to set label to increase customizability
## Problem
The library doesn't have capability to customize label using angular directive.

## Solution
Exposing variables that store label strings on the template's scope, rather than setting with `$.html()`
Here's the [quick simple code change](https://github.com/liooo/angularjs-slider/commit/1b2ccbe010d66df1a31911ea4e7fc0b45b2db874).

and here's the [jsfiddle](https://jsfiddle.net/apm79kcf/6/). 

i roughly looked through the code, and noticed i still need to fix the following things, in addition to the change above.

1. [the logic to recalcute the dimension](https://github.com/angular-slider/angularjs-slider/blob/master/src/rzslider.js#l764-l774)
2. use css `display:hide` rather than `opacity:0` to hide combo label or model/high label as they switch, in case for example there's ng-click attached on either of the element ([written here](https://github.com/angular-slider/angularjs-slider/blob/master/src/rzslider.js#L1262))
3. leaving the backward compatibility (but probably we don't need `translate` option in the future if we're doing this)

## Question
what do you guys think of this idea and the solution?

Liooo:
Gonna fix #355


ValentinH:
I just checkout your code and first reverse the logic about the `inject-label` option (as mentioned in the issue). Then, I commented your watch block and didn't see any issues with the combo label position (even without a $timeout) call.

Here is my translateFn method:
```js
translateFn: function(value, label, which, useCustomTr) {
        useCustomTr = useCustomTr === undefined ? true : useCustomTr;

        var valStr = '',
          getDimension = false,
          dontInjectLabel = label.hasClass('dont-inject-label')

        if (useCustomTr) {
          if (this.options.stepsArray)
            value = this.getStepValue(value);
          valStr = String(this.customTrFn(value, this.options.id, which));
        }
        else {
          valStr = String(value)
        }

        if (label.rzsv === undefined || label.rzsv.length !== valStr.length || (label.rzsv.length > 0 && label.rzsd === 0)) {
          getDimension = true;
          label.rzsv = valStr;
        }

        if(!dontInjectLabel)
          label.html(valStr);
        this.scope[which + 'Label'] = valStr;

        // Update width only when length of the label have changed
        if (getDimension) {
          this.getDimension(label);
        }
      }
```

Can you try it and tell me if you can still get the positioning problem?

Liooo:
@ValentinH Did you set the label using angular binding, like ``{{}}`` or `ng-bind` inside the custom template?

ValentinH:
Nope I just use the code you added in the demo with the custom directive.

Liooo:
@ValentinH
Yea it worked. Probably the problem was somehow caused by ng-bind-html.

README.md
```diff
@@ -1,4 +1,4 @@
-## AngularJS slider directive with no external dependencies
+e# AngularJS slider directive with no external dependencies
```
ValentinH: You probably modify this line by mistake ;)

demo/index.html
```diff
@@ -158,6 +158,16 @@
   </article>
 
   <article>
+    <h2>Slider with angular directive inside custom template</h2>
+    <rzslider
+            rz-slider-model="slider_custom_directive_inside_template.minValue"
+            rz-slider-high="slider_custom_directive_inside_template.maxValue"
+            rz-slider-options="slider_custom_directive_inside_template.options"
+            rz-slider-tpl-url="directiveInCustomTemplate.html"
```
ValentinH: I don't see the directiveInCustomTemplate.html template, did you forget to include it?

Liooo: I forgot this branch was pullreqeusted one, it's just wip commit. Sorry for bothering

ValentinH: No problem! ;)

ValentinH:
:+1:  Great job man!  It's the best PR I've seen since I maintain this project, everything is here: documentation, tests, etc.

Thanks! :)

Liooo:
@ValentinH 
Can you chekc the pullrequest?

If it's looking good, I'll send one for gh-pages as well.

Liooo:
@ValentinH Oh that's nice, thanks man  :)

PR #360, opened 2016-06-30
refactor(AMD/CommonJS exported module): Export module name

### LINKED ISSUE #326: Webpack error
### Steps to reproduce
1. Install webpack
2. Install slider
3. Require slider
4. Inject slider into the app

### Expected behaviour
Should initiate application

### Actual behaviour
On Inject breaks the application

gaelduplessix:
When the module is imported using AMD/CommonJS, it is expected to expose its name instead of the
Angular module object. This is because angular.module() expects the dependencies as string instead
of actual module objects.

Fix #326 

**Breaking change**: code that relies on the module object to be exported (accessing the name
via `.name` for example) will break, since the name is now directly returned.


ValentinH:
👍

ValentinH:
Released under 5.0 version ;)

gaelduplessix:
Merci Valentin ! ;)

Thanks for being that reactive ! 👍

PR #361, opened 2016-06-30
fix(options legend): fix #359

### LINKED ISSUE #359: Cannot switch from steps array to regular steps with ticks values
### Steps to reproduce
1. Instantiate a slider with a `stepsArray`
2. Update the slider options to instead use regular steps with `showTicksValues` set to true

Demo: http://jsfiddle.net/3jjye1cL/135/

### Expected behaviour
The slider should update to the new settings

### Actual behaviour
`TypeError: Cannot read property '0' of null`, slider is broken

gaelduplessix:
cf: #359


ValentinH:
Thanks

PR #367, opened 2016-07-09
added tick color formatting function

amit777:
Allow for custom formatting of tick colors based on the value of the tick.


ValentinH:
Thanks for the pull request!

ValentinH:
Could you add a test case and fix the conflicts please? Also why did you change the outline behavior for selected handles?

amit777:
Hi, yes, I can take a look.  I'm a bit new to social coding so please bare with me.  I changed the outline because it creates an ugly blue box around the handle otherwise (at least on my Mac and Chrome). Here is a screen shot:
screen shot 2016-07-10 at 1 27 47 pm

amit777:
I see that the branch has conflicts, but I'm not sure how to see what those conflicts are?  Are they due to the minified files?

ValentinH:
I don't know, you should rebase on top of master (check the Contributing guide)

PR #371, opened 2016-07-12
feat(directive): Add function to customize color of ticks

amit777:
Pass a custom function to the slider options to allow for dynamic tick color determination

none


ValentinH:
This one is correct, but please remove the change about outline...

amit777:
I can remove change on outline and keep it in my fork.  Just curious if there a specific reason you don't want it merged?

ValentinH:
Outline can be disabled by setting keyboardSupport to false. The outline is here to know which handle is selected and can be moved with keyboard.

Also your pull request is missing a test case for the new feature. You can copy the getPointerColor tests to cover your changes. ;)

amit777:
got it.. i'll add the tests and remove the outline change

Issue #353, opened 2016-06-20
Slider value in view gets disappeared


YuvPrime:
### Steps to reproduce
1. `<div ng-controller="TestController as vm">
        <p>Price: {{vm.priceSlider.value}}</p>
        <div class="col-md-12">
        <div class="col-md-3">
        <rzslider rz-slider-model="vm.priceSlider.value" rz-slider-options="vm.priceSlider.options"></rzslider>
        </div>
    </div>
    </div>`

2. `vm.priceSlider = {
        value: 23,
        options: {
            floor: 20,
            ceil: 100
        }
    }`

Demo: http://jsfiddle.net/cwhgLcjv/87/

### Expected behaviour
At point 23 & 93, the slider min and max value in view should not disappeared.

### Actual behaviour
At point 23 & 93, the slider value gets disappeared. At point 23, min value disappears, at point 93, the end i.e max value gets disappera.

ValentinH:
The min or max labels disappears to prevent them for overlapping the value label.

trodi:
If you choose to hide the value labels, the min/max overlapping isn't a concern anymore. Shouldn't they not disappear in this instance?
`hidePointerLabels: true`

ValentinH:
Indeed, you're right. In this case, it's a bug.

ValentinH:
Fix released in 5.4.1. Thansk ;)

PR #373, opened 2016-07-14
fix(options): Fix showing limit labels when pointer labels are always hidden

trodi:
Previously limit labels would be hidden based on the position of the pointer labels even if user
options hide pointer labels.

This addresses #353


ValentinH:
Thanks for this PR, it looks good. 

I'm wondering if, when the hidePointerLabels is toggled at runtime, everything will work as expected. Indeed, if the option is off and the floor/ceil label is hidden because the pointer is next to it, and the option is turned on, the floor/ceil label will stay hidden while they should be visible...

ValentinH:
I'm away from computer until the end of the weekend so I'll try this scenario later. ;)

trodi:
Here's a [JSFiddle](http://jsfiddle.net/dqgn9oLe/) which should demonstrate your use case. I believe it handles as expected. Below are concrete steps to test (as I understood them).

1. `hidePointerLabels: false`
2. move point to force min limit to hide
3. programmatically toggle  `hidePointerLabels: true`
4. min label is visible and pointer label is hidden

ValentinH:
Indeed, this corresponds to what I was talking about. Good job! ;)

PR #376, opened 2016-07-25
less to sass

saunierpurjus:
converting less files to sass


ValentinH:
Thanks for  taking the time to submit this PR. However, as I've already closed [similar PR](https://github.com/angular-slider/angularjs-slider/pulls?utf8=%E2%9C%93&q=is%3Apr%20sass), I don't want to merge this.

"I don't want having to maintain 2 codebases. If one day someones find an issue with LESS that can be solved by switching to SASS then I will switch to it."

saunierpurjus:
ok thanks

PR #385, opened 2016-08-06
Remove focus border

thierryc:
Remove focus border on webkit browser (Safari and Chrome).
Just a little aethstetic fix.

Smart directive for angular.

Thank you
Thierry

screen shot 2016-08-06 at 23 39 39
screen shot 2016-08-06 at 23 39 06


ValentinH:
The outline is here in purpose for the keyboard support. If you don't want it just disable the keyboard support.

thierryc:
Interesting features. I understand the concept. 
For UX the dark blue dot is good focus.
The default border is too much. The squared focus do not match the circle.

Both options should not be linked. 

This is just a suggestion.
Google Material design follow this option.

Best

PR #405, opened 2016-08-25
Added the option for having fixed label to the directive

dassaptorshi:
If possible please merge it to 5.4.3 as we have started using that version of yours.

Please kindly approve


ValentinH:
Please describe the purpose of this PR and provide a demo.

dassaptorshi:
This PR is to configure the directive so the limit labels don't hide when we reach a particular limit towards floor or ceil value.

[Demo](http://plnkr.co/edit/nOf5YX?p=info)

dassaptorshi:
Hi Valentin,
How long do you think it may take to  merge the pull request?

Thanks in advance

src/rzslider.js
```diff
@@ -1093,8 +1096,10 @@
             backgroundColor: pointercolor
           };
         }
-
-        this.shFloorCeil();
+        if(!this.optionsshowLimitLabels){
```
ValentinH: this is wrong

dist/rzslider.js
```diff
@@ -1089,8 +1092,11 @@
             backgroundColor: pointercolor
           };
         }
-
-        this.shFloorCeil();
+        console.log(this.options.showLimitLabels)
```
ValentinH: this should be removed.

ValentinH:
I'm sorry, I'm currently on holidays so I don't have access to my computer and won't be able to release it anyway...

Here are a few remarks:
-  I would prefer to replace the `showLimitLabels` by something like `autoHideLabels: true` by default so its purpose is more obvious.
- other comments are on the lines directly...

dassaptorshi:
Done!

ValentinH:
The dist file is still incorrect, you should run grunt before commiting...

dassaptorshi:
Done!

ValentinH:
Merged and release under 5.5.0. I have just rename the new option to `autoHideLimitLabels` since it makes more sense.

PR #415, opened 2016-09-22
fix(keyboardSupport): Prevent losing focus when slider is rerendered

### LINKED ISSUE #411: Focus is lost when showSelectionBarFromValue-option changes
### Steps to reproduce
1. Make a slider with a set `showSelectionBarFromValue`.
2. Add a line that changes that attribute in a watcher which watches the value of the slider.
3. Focus the slider and use the keyboard to manipulate the slider.

### Expected behaviour
Changing the `showSelectionBarFromValue` (or any other options this might be a problem for) should not cause loss of focus for the slider.

### Actual behaviour
The watcher triggers and focus is lost from the slider when the `showSelectionBarFromValue` is changed. One has to focus something else and the refocus the slider to regain focus. But changing it again causes loss of focus.

ValentinH:
Fix #411 

Broken demo: http://jsfiddle.net/nqbcL6k1/
Fixed demo: http://jsfiddle.net/pvtotaw2/


ValentinH:
@dagerikhl Can I have your feedback pls?

dagerikhl:
I can confirm that the proposed solution fixes the problem seamlessly, thanks a lot!
The issue seems to have completely vanished, and the fix is as far as I can see optimal.

### In response to your comment on the issue: https://github.com/angular-slider/angularjs-slider/issues/411#issuecomment-248974465
You are correct in that it did not *completely* lose focus. It still appeared focused, and kept the blue square focus-indicator, but it could not be moved by using keys until it was purposely deselected and reselected.

This was on **Google Chrome 53.0.2785.116 m**.

Once again, thanks for being so helpful!

ValentinH:
Great, then I merge it and will bump a new version.

PR #432, opened 2016-10-18
Wrong properties in documentation: min/max -> minValue/maxValue

adripanico:


ValentinH:
The documentation is correct. There is no mandatory naming for the min and max values, it only depends on your html binding. In the README, the declaration is like this:

```html
<rzslider
    rz-slider-model="slider.min"
    rz-slider-high="slider.max"
    rz-slider-options="slider.options"></rzslider>
```

ValentinH:
By the way the min, max and options variables don't necessarily need to be stored in the same object. It just cleaner...

Issue #409, opened 2016-08-31
Support for `template` not just `templateUrl`


mateatslc:
### Steps to reproduce

N/A
### Expected behaviour

Would be nice to have a new `rz-slider-tpl` directive attribute instead/along with the current `rz-slider-tpl-url`. That would help folks using webpack to require their custom templates upfront as usual, without the need to lube up `$templateCache` upfront.
So with something like this in a controller:

```
scope.tplt = require('./my_custom_slider_template.html');
```

... markup would look like this:

```
<rzslider rz-slider-tpl="{{ tplt }}" rz-slider-model="slider.value" rz-slider-options="slider.options">/rzslider>
```
### Actual behaviour

Only `rz-slider-tpl-url` is supported.

ValentinH:
I don't see any reason not to add it. Can you submit a PR for this?

mateatslc:
I'll try to, as my time allows. 👍

ValentinH:
It's up to you, I'm currently on holidays and don't have access to a computer...

kevin-king:
Just encountered a related issue. From source code

```
templateUrl: function(elem, attrs) {
  //noinspection JSUnresolvedVariable
  return attrs.rzSliderTplUrl || 'rzSliderTpl.html';
},
```

in order to use a template url string from a controller, you need to use rzSliderTplUrl="ctrl.templateUrlValue". Was trying rz-slider-tpl-url="ctrl.templateUrlValue" for hours, but the case conversion of the attribute only occurs when looking off the directive's scope: {}, but apparently not when accessing it directly from attrs like above.

ValentinH:
@kevin-king you are probably doing something wrong because there is a working example on the [demo website](http://angular-slider.github.io/angularjs-slider/): "Slider with custom template to use HTML formatting for ticks"

kevin-king:
My case is a bit different from that one. Here would be an example of the issue I ran into, using the same syntax as your site's examples:

JS
==
```
//You need to import the ngSanitize module:
angular.module('app', ['rzModule', 'ngSanitize']);

$scope.templateUrl = 'your-custom-template.html';

$scope.slider = {
  value: 100,
  options: {
    floor: 0,
    ceil: 500,
    step: 100,
    showTicksValues: true,
    translate: function(value) {
      return '<b>Price:</b> $' + value;
    }
  }
};
```

HTML that wont work
==============
<rzslider rz-slider-model="slider.value"
[...]

PR #438, opened 2016-10-25
Update to templateUrl to allow compilable string values

haxxxton:
As per [issue 409](https://github.com/angular-slider/angularjs-slider/issues/409) there is a request for custom values to be allowed as part of the templateUrl parameter.

ie.

``` html
<rzslider
      rz-slider-model="slider_custom_template.minValue"
      rz-slider-high="slider_custom_template.maxValue"
      rz-slider-options="slider_custom_template.options"
      rz-slider-tpl-url="{{ vm.userSelectedOption }}/rzSliderTemplate.html"
    ></rzslider>
```

Unfortunately, the `templateUrl` directive option occurs before the scope compiles the double brackets, as such loading of the template has been moved within the link option. Because of the nature of this method, it has meant that the slider contents are now one level deeper (as the template html is loaded inside the directive element). This has meant the addition of a container class and style for the slider element

I have not added additional tests or documentation for this upgrade, however all current tests are passing.


ValentinH:
To be honest, I really don't like this solution. It feels more like a workaround than something that should be integrated in the library (having the load yourself the template , compile and attach the content to the DOM).

Thus, I would prefer to keep this solution in a fork.

haxxxton:
No problem, if there is other solutions for compiling custom templates on the fly i am open to exploring them, as ive had to use this method before in a couple of other projects.

Would a solution that uses `ng-include` feel nicer to you? ([Example](http://stackoverflow.com/a/21835866/648350))

ValentinH:
I have try that but it doesn't work since the slider needs the template content to initialize properly. Did you manage to make it work?

haxxxton:
I think as long as you wait until after the next digest cycle, ie. after a 0 length `$timeout`. The template should be included into the directive and you could initialise the slider. I will have a look over the weekend :)

PR #439, opened 2016-10-25
ui.bootstrap upgrade

haxxxton:
Update to the demo html and js files to use latest ui.bootstrap component names, namely

`$modal` -> `$uibModal`

`<tab>` -> `<uib-tab>`
`<tabset>` -> `<uib-tabset>`

PR #441, opened 2016-10-26
update to use latest ui.bootstrap components

haxxxton:
[As suggested](https://github.com/angular-slider/angularjs-slider/pull/439#issuecomment-256261333) in development ui.bootstrap upgrade pull request, i have updated the gh-pages branch.

Im unfamiliar with the structure of gh-pages, but believe i have found all instances of the old component names

### DIFF
--- a/demo.js
+++ b/demo.js
@@ -24,7 +24,7 @@ app.directive('clickableLabel', function () {
   };
 });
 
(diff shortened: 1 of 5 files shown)


ValentinH:
Looks good to me! :+1: thanks again :)

PR #449, opened 2016-11-03
add grunt task that copies the final css to scss file

ronnyrin:
I know that you don't want to maintain two codebases. With this solution, you won't have to.
We need a way to import the final css in our scss file and currently importing css is not supported.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -147,6 +147,14 @@ module.exports = function(grunt) {
         configFile: 'karma.conf.js',
         singleRun: true
       }
+    },
+
+    copy: {
+      copyToSass: {
+        files: [
+          {expand: false, src: ['dist/rzslider.css'], dest: 'dist/rzslider.scss'},
+        ]
+      }
     }
   });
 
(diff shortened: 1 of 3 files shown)


ronnyrin:
Thanks, Can you bump the version?

ValentinH:
Done -> 5.8.3

ronnyrin:
Maybe we can just rename the scss file so there will be no conflict. Importing raw CSS didn't work for me.

ValentinH:
Are you using a recent version of SASS like explain in the linked issue?

ronnyrin:
Yes. It doesn't work.. I'm using [grunt-contrib-compass](https://github.com/gruntjs/grunt-contrib-compass)
```
 error app/styles/main.scss (Line 18 of app/styles/components/...: File to import not found or unreadable: angularjs-slider/dist/rzslider.
Load paths:
  /Users/..../app/bower_components
```

```
✗ ./node_modules/node-sass/bin/node-sass -v
node-sass       3.10.1  (Wrapper)       [JavaScript]
libsass         3.3.6   (Sass Compiler) [C/C++]```

ValentinH:
It seems that you are missing the .css part of your file pat, don't you?

ronnyrin:
No, if you add the .css suffix then it just [transform](https://github.com/sass/libsass/pull/754#issuecomment-96062994) it to `@import url(file.css);` and that won't work at all because the browser won't look in the bower folder.

PR #454, opened 2016-11-08
Add Math.round for positions and dimensions

DmitryKrekota:
By working with this library angularjs-slider library I ran into a problem, that some positions and dimensions sometimes set as 111.333px (not integer value) and browsers render my .svg images differently (sometimes blurry). See the screenshot:
https://drive.google.com/file/d/0Bxtv10NlSOEbM1d5cHZ4VnJwTWM/view?usp=sharing
So, I suggest to add Math.round calls in some places. The decision forces browsers work equally and solves my problems.


ValentinH:
Thanks for contributing this, I would be happy to merge it. Before, you need to fix the unit tests and generate the dist files. You can read the CONTRIBUTING.md file to help you if you want.

DmitryKrekota:
Done. I did all that you asked of me. Please check difference.

tests/specs/options-handling-test.js
```diff
@@ -1217,8 +1217,8 @@
         };
         helper.createRangeSlider(sliderConf);
 
-        var expectedDimension = Math.floor(helper.slider.valueToPosition(6)),
-          actualDimension = Math.floor(helper.slider.selBar[0].getBoundingClientRect().width);
+        var expectedDimension = helper.slider.valueToPosition(6),
+          actualDimension = helper.slider.selBar[0].getBoundingClientRect().width + 1;
```
ValentinH: Why do you need to add 1? It seems weird...

DmitryKrekota:
I removed Math.round from valueToPosition method, because of weird code in tests. So now, after obtaining value, you decide by yourself use Math.round or Math.floor. Please check again.

ValentinH:
Thanks for this. `Math.floor` was a trick to handle decimal value but it should not be useful anymore thanks to your update in code.

Do the tests pass if you put `Math.round` in `valueToPosition` and remove it from the `updateTicks` one? Indeed, it would be cleaner IMO

DmitryKrekota:
> Do the tests pass if you put Math.round in valueToPosition and remove it from the updateTicks one?

No, some tests fail. Therefore, I propose to leave as is.

ValentinH:
Thanks for rebasing.

ValentinH:
Released via 5.8.7

DmitryKrekota:
Thanks your activity

Issue #453, opened 2016-11-07
maxRange option not applied with pushRange


GuilloOme:
Demo: http://jsfiddle.net/sh49nvLs/

### Steps to reproduce
move the right slider to the right: in one case, it's "blocked" in the other case, nothing append.

### Expected behaviour
We would expect the other handle to follow the first one.

ValentinH:
Well spotted, this is a bug...

GuilloOme:
It looks like it's missing some `this.options.maxRange` logic in [`applyPushRange()`](https://github.com/angular-slider/angularjs-slider/blob/master/src/rzslider.js#L2146)

ValentinH:
Actually, I think that `applyMinMaxRange` should be called after [this line](https://github.com/angular-slider/angularjs-slider/blob/master/src/rzslider.js#L2062).

ValentinH:
Fixed in 5.8.6

PR #456, opened 2016-11-08
fix(applyPushRange): Apply the pushRange with maxRange

GuilloOme:
Apply the pushRange when the difference is greater than the maxRange set

Close issue #453

Fixed demo: http://jsfiddle.net/nx7yqqa5


ValentinH:
Thanks for this PR, it is just perfect! It feels good to have such a nice contribution with tests and everything ;)

Actually your solution is better than the one I suggested in the issue because mine would have just blocked the slider when maxRange would be reached while your "pull" behavior is neat!

ValentinH:
Released under 5.8.6

GuilloOme:
Thank you ! Glad to help…

PR #473, opened 2016-12-07
feat: Add selectionBarGradient option to customize the selection bar

Masadow:
Add selectionBarGradient option which create a linear gradient for the selection bar. The option

accepts an object that contains `from` and `to` properties which are colors describing the gradient

Here's a fiddle demonstrating the feature: http://jsfiddle.net/masadow/cqk61f0a/

Additional note: There's room for improvement, for example, the ticks are not affected by the gradient. I think it's another feature but if someone feels like it's a good idea, feel free to give it a go.


ValentinH:
Why not just doing it via CSS?

Masadow:
It's not possible with the range slider. Or I don't see how without changing the whole DOM.

Btw, it seems like a test must be written to hit 100% of code coverage ?

ValentinH:
Ok I understand now. Thanks for this pretty complete pull request. Indeed, you need to add tests to cover your changes. You can start from an existing one related to the selection bar. ;)

dist/rzslider.js
```diff
@@ -1268,6 +1269,20 @@
           this.scope.barStyle = {
             backgroundColor: color
           };
+        } else if (this.options.selectionBarGradient) {
+          var offset = this.options.showSelectionBarFromValue !== null ? this.valueToPosition(this.options.showSelectionBarFromValue) : 0,
+            reversed = offset - position > 0 ^ isSelectionBarFromRight,
```
Masadow: It's a XOR. It means either `offset - position > 0` or `isSelectionBarFromRight` is true, but not both.

Masadow:
I'll write some test within the day and update my PR. Can I add another commit or should I `git reset --soft HEAD~1` my current commit and create a new one ?

ValentinH:
Adding another commit is OK ;)

dist/rzslider.js
```diff
@@ -1268,6 +1269,20 @@
...
+            direction = this.options.vertical ? 'top' : (reversed ? 'left' : 'right');
```
ValentinH: the vertical + reversed case is not handled I think. Look this demo: http://jsfiddle.net/nLk7p00v/

Masadow: Indeed, I did not see pay attention to that option `showSelectionBarEnd`. It's fixed by now

ValentinH:
The vertical reversed example is still broken: 
image
 
There is a white part on top...

Masadow:
I made a mistake in the fiddle. I was not pointing to the right branch, please try with `<script src="https://cdn.rawgit.com/Masadow/angularjs-slider/gradient-selection-bar/dist/rzslider.min.js"></script>`

ValentinH:
OK you are right. I think it is OK to merge it, I'll just need you to rebase on top of master to resolve the current conflicts (just the dist files).

For later, I just point out that the `showSelectionBarFromValue` doesn't work with the gradient but it'll be more complex to handle it I think: http://jsfiddle.net/o0x97L60/

Masadow:
That's fine for conflicts, I rebased on master already.

I agree about showSelectionBarFromValue, I partially handled it but I did not calculate the length of the background gradient. I thought I would not try to handle it since there are multiple design decisions to go from

ValentinH:
Indeed, I don't even know how I would handle it, so that's fine for now.

ValentinH:
Thanks for your contribution! :+1:

ValentinH:
Released via 5.9.0

PR #474, opened 2016-12-08
prevent angular being loaded twice when using with browserify

nickccm1122:
The console shows a warning that angular is being loaded twice as angular is loaded as global. 

Fixed to use global angular object if it is defined.


ValentinH:
Thanks for this. I'll merge it later today (so I can bump a new npm version at the same time) if I have time.

ValentinH:
Released under 5.8.8

PR #501, opened 2017-02-15
Range outside

alexpireddu:
Enable the user to have maxValue < minValue, highlighting the space around instead of the space between.
Added a new option to support this: showOutRange


ValentinH:
Hi Thanks for contributing to this project.
Could you provide a JSFiddle showing what does this PR please?

alexpireddu:
Hi,
I put a little demo here: https://jsfiddle.net/ekLbqcwr/8/
there is one also in the demo folder.

alexpireddu:
Hi ValentinH,
I improved a couple of things more for this case but I need to introduce a breaking change.
I push on my fork, I close this and open a new pull request?

ValentinH:
I'm sorry but I don't understand the point of this feature... Could you explain a bit more why you need this?

alexpireddu:
I'm actually using the slider to show a range of time during the day.
You can have for example 9->18 or 18->9.

ValentinH:
OK, now I get it. What would be the breaking change?

alexpireddu:
It's for who is using the custom template, they'll need to add another rz-bar-wrapper in the same position as here (line 6,7,8):
https://github.com/alexpireddu/angularjs-slider/blob/be655f10dbb945caaae33640d0116c7636d44140/tests/specs/custom-tpl.html

ValentinH:
To be honest, I don't want to add this to the library because it's adding quite a bit of complexity just for a really specific use case. I would prefer that you keep this in a fork.

alexpireddu:
Yeah probably it suits good only for the time case.

ValentinH:
Indeed and also there are many edge cases to handle with all the options that you probably don't need.. Thus, you can create you custom version with only your requirements.

PR #502, opened 2017-02-16
Update the labels on separation

ku:
When mergeRangeLabelsIfSame is set to true, two labels are combined once their values got close.
They got separated again if the values got away enough, but the text of a label stays the same with old one which is shown when it was combined.

Following gif shows the problem described above. 
timeslot

This patch solves the problem by updating the labels on separating combined labels.


ValentinH:
Sorry for the delay, I haven't found the time to merge it for now (I want to merge and release it in the same time that's why). I'll try to do it soon. :smile:

PR #518, opened 2017-03-29
Vendor prefixes for transform property

bod1ka:
Added several vendor prefixes for transform property to fix vertical ticks placement on the browsers that require vendor prefixes


ValentinH:
Thanks for this! :tada:

ValentinH:
Released under 6.1.1.

I see don't see the prefixes in the inline style on Chrome, did this fix your issue on safari?

bod1ka:
Yes , definitely.

PR #531, opened 2017-05-14
changed top: initial to top: auto for ie support

anuragasaurus:
Changed `top: intial` to `top: auto` . Works fine in IE11, chrome and mozilla.


ValentinH:
Thanks for this? I'll merge it tonight.

Issue #534, opened 2017-05-23
Imposible to use many sliders in the same time on a device with touch screen (by multitouch)


daniela-mateescu:
### Steps to reproduce
1. Having a device with a touch screen
2. Having two sliders on the same screen
3. Start dragging first slider 
4. Start dragging second slider in the same time using another finger
5. Continue dragging both sliders in the same time
Demo: http://jsfiddle.net/1o71k7j5/1/

### Expected behaviour
Both sliders are dragged correctly, according to the corresponding finger that initiated the drag

### Actual behaviour
Only the second touched slider is moving when moving both fingers. Also this slider moves incorrectly because it follows the movement of the finger that initiated the drag on the other slider.

### Notes
We already done some internal test by extending your component and adding the fix to the above issue. We want to create a pull request for this change. But before starting to integrate this fix on your library we would like to know if you think that it would be ok to integrate it? 

Thank you and looking forward for your response.

ValentinH:
Hello. Any pull requests are always welcome. The only conditions are:
- it does not break other features
- it keeps the test coverage at 100%

Thus, I would be more than happy to provide any useful information if you need. ;)

PR #535, opened 2017-05-25
fix: Fix for wrong slider drag when having many touches simultaneous

daniela-mateescu:
#534 
Sample at: http://jsfiddle.net/fq6xkod4/ . This sample requires a device with a touch screen. It illustrates that the possibility to use both sliders simultaneously.


ValentinH:
Thanks for this PR. It is really clean and provide full test coverage! Thanks 👍

daniela-mateescu:
Hello! Thank you very much for integrating the PR, but we have a little problem with this. I saw that you did a refactor before integration. It is all ok with this refactor, except one little problem. At onStart() method, instead of memorizing the  ehEnd handler to be removed later like this: ` this.ehEndToBeRemovedOnEnd = ehEnd;`, you have sent this as a parameters like this: ` ehEnd = angular.bind(this, this.onEnd, ehMove, ehEnd);`. The problem with this code is that the `ehEnd` variable will be always undefined when passing it to the `angular.bind` method call.

Thank you in advance.

ValentinH:
I modified after merging yesterday because of this issue #536 that was related to this PR. But you are right, I did it a bit too quickly and it's wrong... 

It proves that even with 100% test coverage, some tests are still missing because this should have been caught. I'm fixing it right away by keeping the way you were doing it but also by providing the correct event name.

ValentinH:
Master should be fixed on http://jsfiddle.net/mejsy5d2/

daniela-mateescu:
Hello,
Thank you very much for the fix. It works great now.

PR #547, opened 2017-07-26
Adding a new option showSelectionBarSingle to allow getting the rz-se…

asafblum:
…lected on a single tick


asafblum:
image

ValentinH:
Thanks for this PR. However, I think that the option name is not understandable. What does it really do? Could you provide a demo using JSFiddle?

Also this is breaking the test coverage, so a new unit test is needed.

asafblum:
Hi @ValentinH please take a look: http://jsfiddle.net/asafblum/wtboevs4/1/
also added a test does this option make sense or I can achieve this with out the new option:
I'd like to get the selected class on the exact tick for the current value and not from

ValentinH:
Thanks for the example. I'm starting to wonder if an option is needed and if it should just be the default behavior to add this class to the tick. ;)

asafblum:
Hey @ValentinH ,
What's the direction on this one? can I close the PR?

will you use it?

ValentinH:
I think we should not add a new option, but fix the current behavior by always setting `.rz-selected` on a single slider current value tick.

PR #563, opened 2017-08-30
No background when pointer is selected

dfavaro:


ValentinH:
Thanks for the pull request. However, I don't understand what it does... Can you provide a description and a demo?

PR #567, opened 2017-09-07
Added reversedControls option to reverse keyboard navigation.

DadUndead:
It's a handy option if you want to reverse controls of your slider. I needed such for my project.


ValentinH:
I don't really why this is needed, but since your PR is perfect and don't add any overhead, I just merge it. Thanks for it! 😉

PR #571, opened 2017-09-18
Call the onChange event when model is updated

gilluminate:
Currently the onChange even only fires if there is mouse or keyboard actions happening. This change also fires the event if the model changes outside the directive which gets watched and updated.

### DIFF
--- a/src/rzslider.js
+++ b/src/rzslider.js
@@ -424,6 +424,7 @@
             if (self.internalChange) return
             if (newValue === oldValue) return
             thrLow()
+            self.callOnChange()
           })
 
           this.scope.$watch('rzSliderHigh', function(newValue, oldValue) {


ValentinH:
Hello @gilluminate. Sorry for not answering before, but I must have missed the notification back then :/

Regarding this PR, I think I wanted to do the same thing in the past but it can lead to issues. The goal of the onChange is to notify the parent when the slider has changed its model. In this case, this is not the case and the parent definitely knows when the model changes so it can just call the same function that you would attach to the onChange listener.

gilluminate:
I see what you are saying. The value gets set outside of the component and then the onChange gets called outside of the component, so the controller instantiating the component should be able to handle this just fine.

PR #572, opened 2017-09-21
Update rzSliderTpl.html

alexmironof:
Fix unreal lags when operate billion numbers range


ValentinH:
Hi @alexmironof , thanks for this pull-request!

Could you set up an example showing the issue you mention? You can fork this fiddle: http://jsfiddle.net/cwhgLcjv/

ValentinH:
Also, the unit tests are broken with your change.

alexmironof:
@ValentinH it was very hasty decision, I think the problem is that I am using slider package from bower registry :sweat_smile: 

This example works perfect on [JSFiddle](http://jsfiddle.net/s0xe23ao/)

alexmironof:
@ValentinH bower is broken: It installs version 6.4.0 (I saw it in package.json), but in dist folder I've got version 6.3.0

PR #590, opened 2017-11-14
fix(options): react to changes of options of Function type

### LINKED ISSUE #589: Changing options of Function type in runtime is not supported
### Steps to reproduce
1. try changing any option which type is Function in runtime

Demo: http://jsfiddle.net/95dLpcgs/

### Expected behaviour
New option is applied - as stated in docs

### Actual behaviour
Option is not applied

### Investigation result

This is because when watching on options under the hood `angular.equals` is used, which ignores functions - try running `angular.equals({a: () => 1}, {a: () => 2})` - it returns `true` though objects aren't equal (I assume this is by design)

### Possible workaround

Every time when Function option is changed - add some random value to options (ugly, but works - look at demo page above)

zuzusik:
fixes #589


ivanvoznyakovsky:
great find 💯

ValentinH:
Good stuff!

ValentinH:
Released under 6.4.1

Issue #597, opened 2017-11-21
TypeError: (intermediate value)(...) is not a function


mlatzko:
I use the angularjs-slider together with other 3rd party JS code an bundle it into one file. Because since version 6.4.1 the trailing semicolon is missing at the end of the closure, in `src/rzslider.js`, which causes an type error when bundled together with other libraries.

e.g. `(function(){})()(function(){})()` Error
e.g. `(function(){})();(function(){})()` No error

bilalhammoud:
since yesterday i am running into this issue as well, could you fix it please as soon as possible

logan-jobzmall:
The workaround I found is to make sure to terminate lines 2661 and 2662 with ';' in the dist/rzslider.js. Don't know how that computes in the main src. But it should help.

mlatzko:
> Moreover, using https://github.com/angular-slider/angularjs-slider/blob/master/dist/rzslider.min.js should be fine since the last char is the semicolon.

As @ValentinH points out in the closed pull request #598 I now use the mininfied version. This solves the problem. Thx @ValentinH

PR #598, opened 2017-11-21
#597 fix issue 

mlatzko:
Fix issue causing type error while bundled with other js code.

### DIFF
--- a/src/rzslider.js
+++ b/src/rzslider.js
@@ -2656,4 +2656,4 @@
   /*templateReplacement*/
 
   return module.name
-})
+});


ValentinH:
Do you need this if you use the `dist` files?

mlatzko:
Yes I would need this in the `dist` files.

But the the `src` files need them as well, cause the missing semicolon should raise a jshint warning as this little example demonstrates.

Just copy/paste to http://jshint.com/.
```
(function(root, factory){
  "use strict";
  

})(this, function(angular){
	"use strict";
  
})
```
Result: 1 Warning - Line 8 - Missing semicolon.

mlatzko:
Any progress, we can either close the pull-request if there is a way to add the needed semicolon in the `dist` files or we can apply the semicolon in `src` files both solution are fine with me. 

Let me know if I can help out.

ValentinH:
The thing is that you can see this issue the other way around: we already add a `;` at the beginning of the code (https://github.com/angular-slider/angularjs-slider/blob/master/src/rzslider.js#L11), thus the issue can also be seen as coming from the library that is appended after this one in your concatenated file.

We are using a really standard tool, aka Prettier, without any semicolons in the source files (except from the first mandatory one). The issue might be coming from your packaging system as well.

ValentinH:
Moreover, using https://github.com/angular-slider/angularjs-slider/blob/master/dist/rzslider.min.js should be fine since the last char is the semicolon.

mlatzko:
> Moreover, using https://github.com/angular-slider/angularjs-slider/blob/master/dist/rzslider.min.js should be fine since the last char is the semicolon.

Thx for pointing this out and sorry for the inconvenience I now use the minified version in my build process.

ValentinH:
No problem ;)

PR #612, opened 2018-01-22
fix(onStart): Remove stop propagation call so events will bubble up

### LINKED ISSUE #611: Removing stopPropagation
Hello, I'm using the slider inside a sweetalert2 modal and have a problem with users trying to slide to the min or max value and accidentally closing the modal.

Clicking and sliding to min/max value with the mouse going outside the modal then releasing the click will close the modal because of click event on background overlay.

Removing stopPropagation fixes the issue because sweetalert2/boostrap knows if the background overlay click was intended or not based on mousedown/mouseup.

Made some tests with stopPropagation removed and seems to work correctly. Will a pull request with this change be merged?

https://css-tricks.com/dangers-stopping-event-propagation/

### Expected behaviour
The mousedown events should propagate upwards.

### Actual behaviour
The mousedown events don't propagate because stopPropagation is called.

andreeib:
Closes #611


ValentinH:
Thanks! :)

stripathix:
In ionic, there is an event attached to swipe left/right to show/hide side menu. Removing event.stopPropagation() on start causes opening of side menu when the slider is dragged to right. There can be many different cases or framework which has an event attached to swipe especially for mobile application, I am not sure removing event.stopPropagation() from the library was a good idea because now every developer who uses this library for mobile apps will have to write onStart option to ensure event.stopPropagation(); to prevent side menu from opening. 

To solve this I have added an option for slider configuration. 

`options: {
                onStart: function () {
                    event.stopPropagation();
                },
}`

stripathix:
Adding onStart as an option does not solve this issue of side menu opening. Rolling back to 6.4.3 :-(

PR #616, opened 2018-02-06
feat($compile): adds labelOverlapSeparator as an option

ben-henoida:
The labelOverlapSeparator option allows the user to specify what string to use to separate the
labels when they overlap. The previous hardcoded string ' - ' could indeed be mistaken as a minus
sign. To avoid introducing a breaking change, the default is the previous value: ' - '. Tests and
README.md modified accordingly.

Demo: http://jsfiddle.net/sLxkwv9s/
(Compare to original problem: http://jsfiddle.net/04sr3ty0/1/)


ben-henoida:
I have removed the file, would it be a good idea to add it to .gitignore ?

ValentinH:
Indeed, can you add this as well? :)

ValentinH:
Perfect!

ValentinH:
Released! https://github.com/angular-slider/angularjs-slider/releases/tag/6.5.0

Issue #410, opened 2016-09-01
Problems using enforceRange/enforceStep


richardkhchang:
Hi,

I'm currently trying to use this control to create a price range filter.

This is my slider:

```
$scope.slider = {
                minValue: 1,
                maxValue: 10000,
                options: {
                    floor: 1,
                    ceil: 10000,
                    enforceStep: false,
                    enforceRange: false,
                    mergeRangeLabelsIfSame: true,
                    showTicks: true,
                    pushrange: true,
                    stepsArray: [
                        { value: 1 },
                        { value: 10 },
                        { value: 20 },
                        { value: 50 },
                        { value: 100 },
                        { value: 200 },
                        { value: 500 },
                        { value: 1000 },
                        { value: 5000 },
                        { value: 10000 }
                    ]
                }
            };
```

And this is my front-end markup:

```
<rzslider rz-slider-model="slider.minValue"
              rz-slider-high="slider.maxValue"
              rz-slider-options="slider.options"></rzslider>

<label for="minPrice">Min $</label><input type="number" id="minPrice" name="minPrice" ng-model="slider.minValue" ng-model-options="{updateOn: 'blur', allowInvalid: false}" min="{{slider.options.floor}}" max="{{slider.maxValue}}" required>
<label for="maxPrice">Max $</label><input type="number" id="maxPrice" name="maxPrice" ng-model="slider.maxValue" ng-model-options="{updateOn: 'blur', allowInvalid: false}" min="{{slider.minValue}}" required>
```

As you can see, I have both the angularjs-slider as well as two "number" textboxes hooked up to the same model,  "slider.minValue", and "slider.maxValue". When the slider is manipulated, the values in the textbox correctly update to the selected value in "stepsArray". This much is working correctly.

However, I want to allow the user to enter into the textboxes a specific price range that does not perfectly match the values in the "stepsArray". I was hoping that doing this would move the slider to the appropriate position between the defined steps, and that subsequent slider movements would snap the values back in sync with the "stepsArray".

Looking at readme.md, I'm under the impression that the enforceStep or enforceRange properties would allow me to do this by setting one or the other to false.

> enforceStep - Boolean (defaults to true): Set to true to force the value to be rounded to the step, even when modified from the outside.. When set to false, if the model values are modified from outside the slider, they are not rounded and can be between two steps.
> 
[...]

PR #626, opened 2018-05-11
fix slider jumps to first value if value is not in stepsArray

LembDo:
As I used the slider with a stepsArray and an input to change the value, I found that if a value is entered in the input that's not in the stepsArray the slider jumps to its minimum value.

I changed it so that it jumps to the next higher value instead.

### DIFF
--- a/src/rzslider.js
+++ b/src/rzslider.js
@@ -450,7 +450,9 @@
           var index = 0
           for (var i = 0; i < this.options.stepsArray.length; i++) {
             var step = this.options.stepsArray[i]
-            if (step === modelValue) {
+            var lastStep = i > 0 ? this.options.stepsArray[i - 1] : 0;
+
+            if (step === modelValue || (modelValue > lastStep && modelValue < step)) {
               index = i
               break
             } else if (angular.isDate(step)) {


src/rzslider.js
```diff
@@ -450,7 +450,9 @@
           var index = 0
           for (var i = 0; i < this.options.stepsArray.length; i++) {
             var step = this.options.stepsArray[i]
-            if (step === modelValue) {
+            var lastStep = i > 0 ? this.options.stepsArray[i - 1] : 0;
+
+            if (step === modelValue || (modelValue > lastStep && modelValue < step)) {
```
praveeno: some context require here, will you add comments here

DominikFG: If a stepsArray is given in the options and the value is set from the outside to a value that is not part of the stepsArray, currently the slider will jump to the first element in the stepsArray. I changed it so that if the value is between two values the slider jumps to the higher value.

ValentinH: I like this feature. However, could you add a test case for it?

evestraw:
I also run into this issue when the ceil is not in the step interval the ceil could be overshot.

ViieeS:
I think we just need to respect `enforceStep` option when using `stepsArray`. This is old bug #410.

PR #662, opened 2019-02-20
add the option to use ticksArray as array of objects, with legend and value properties

vdiez:
same as we do with steps array, but to customize the ticks without limiting the selection steps. 

I know this can be already done with a variety of methods, but AFAIK only with functions. In my case I store these labels in a mongo db, so storing logic is not ideal. Putting those labels in a table of objects without limiting the selection as stepsArray solves both issues, storing and displaying.


ValentinH:
Hey @vdiez, thanks for this PR!

Could you create a JSFiddle showing how to use this option? For this you can fork the example one from the ReadMe and point the script (angular-slider.js) to your repo instead of this one. You'll need to run the `build` script on your PR to generate the different needed files.

Also, please provide a unit test for this option so we keep the coverage at the same level as before.

Thanks :)

vdiez:
Hi @ValentinH,

fiddle here: http://jsfiddle.net/3jgzv4wc/1/

Cheers

ValentinH:
Released under 6.7.0 🎉

PR #664, opened 2019-03-05
Add 'slideChanged' event

marlais:
By adding a 'slideChanged' event to the callOnChange function, you're giving the user scope a much easier way to detect slider updates than having to listen to the local variable manually. I personally found this useful in my project, but I would rather not modify the source code on your project.

### DIFF
--- a/dist/rzslider.js
+++ b/dist/rzslider.js
@@ -2674,6 +2674,7 @@
               )
             })
           }
+          this.scope.$emit('slideChanged')
         },
 
         /**


ValentinH:
Hello,
Thanks for this PR.

You don't need to listen to the local variable, you can simply use the `onChange` callback. Do you need something else?

ValentinH:
Closing as there was no answer for a while. Comment to re-open 😉

